PR #118, opened 2026-01-29
Add two-tier query pathway for DuckDB-specific syntax

EDsCODE:
## Summary

- Implements automatic fallback to native DuckDB execution when PostgreSQL parsing fails
- Allows DuckDB-specific syntax (DESCRIBE, SUMMARIZE, FROM-first, ATTACH, CREATE SECRET, etc.) to work without explicitly enabling `native_duckdb` mode
- Validates fallback queries using DuckDB's EXPLAIN before execution (skipped for utility commands)
- Handles leading SQL comments correctly

## Changes

- **transpiler/config.go**: Add `FallbackToNative` field to Result struct
- **transpiler/transpiler.go**: Return `FallbackToNative` result instead of error on parse failure
- **server/conn.go**: Add `validateWithDuckDB()` and `isDuckDBUtilityCommand()` helpers, handle fallback in query and extended query protocol paths
- **tests/**: Comprehensive unit and integration tests

## Test plan

- [x] Unit tests for transpiler FallbackToNative behavior
- [x] Integration tests for DuckDB-specific syntax (DESCRIBE, SUMMARIZE, FROM-first, etc.)
- [x] Integration tests for utility commands (ATTACH, DETACH, USE, SET, RESET)
- [x] Integration tests for queries with leading comments
- [x] Verified with dbt running CREATE SECRET and ATTACH commands

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -1221,6 +1221,14 @@ func queryReturnsResults(query string) bool {
 	if strings.HasPrefix(upper, "DESCRIBE") {
 		return true
 	}
+	// SUMMARIZE returns results (DuckDB-specific)
+	if strings.HasPrefix(upper, "SUMMARIZE") {
+		return true
+	}
+	// FROM-first syntax returns results (DuckDB-specific)
+	if strings.HasPrefix(upper, "FROM") {
+		return true
+	}
 	return false
 }
 


fuziontech (commented):
## Review: Two-tier query pathway for DuckDB-specific syntax

Nice feature - this will help dbt users a lot. A few things to consider:

### Critical

**1. Prepared statements with parameters won't work with fallback**

When `FallbackToNative=true`, `result.ParamCount` is 0 (parser never ran). If someone does:
```sql
PREPARE stmt AS DESCRIBE SELECT * FROM users WHERE id = $1
```
Parameter binding will fail confusingly. Consider detecting `$N` placeholders in fallback queries and erroring explicitly, or document this limitation.

**2. EXPLAIN in error messages is confusing**

When `validateWithDuckDB` fails, users see errors like:
```
syntax error: Parser Error: EXPLAIN SELECT * EXCLUDE foo...
```
The "EXPLAIN" prefix is confusing since they didn't run EXPLAIN. Consider stripping it from error messages.

### Medium

**3. `FROM ` prefix bypass is unnecessary**

FROM-first queries support EXPLAIN in DuckDB. Including `"FROM "` in `utilityPrefixes` means invalid FROM-first syntax bypasses validation. Suggest removing it.

**4. Inconsistent trailing spaces in utility prefixes**

`"ATTACH"` vs `"USE "` - inconsistent. A hypothetical `ATTACHDB` command would incorrectly match. Consider consistent word boundary matching.

**5. `TranspileMulti` fallback treats entire batch as one query**

`SELECT 1; DESCRIBE users` fails PostgreSQL parsing on statement 2, but returns the entire batch as a single fallback query. Is this intentional?

**6. `queryReturnsResults` doesn't know about DuckDB commands**

May cause issues with extended query protocol's Describe for commands like `DESCRIBE`, `SUMMARIZE`.

### Minor

- Comment says DESCRIBE supports EXPLAIN "in some cases" but it's still in the skip list - clarify?
- No tests for fallback with parameterized queries (failure mode)
- Consider a `strict_postgresql` config to disable fallback for users wanting strict PG compatibility

---

Overall the implementation is solid and tests are comprehensive. The prepared statement parameter issue is the main concern.

jghoman:
Can we document this behavior in the README.md?

PR #240, opened 2026-02-19
fix: prevent worker socket collisions during handover

EDsCODE:
## Summary

Fixes "too many connections" errors that occur during control plane handover (deploy), even when no client connections are active.

## Root cause

During handover, the new CP creates a fresh `FlightWorkerPool` with `nextWorkerID = 0`. Its workers get socket paths like `worker-0.sock`, `worker-1.sock` — **the same paths the old CP's workers are still using** while draining.

The race plays out like this:

1. Old CP receives SIGUSR1, spawns new CP via self-exec
2. New CP completes handover, takes over the PG listener, starts accepting connections
3. A connection arrives → `AcquireWorker` → `SpawnWorker(0)` → tries to create `worker-0.sock`
4. `SpawnWorker` calls `os.Remove("worker-0.sock")` — deletes the old worker's socket file
5. The new worker process starts and tries to `bind()` on `worker-0.sock`
6. But the old CP's worker-0 process is **still alive** (old CP drains connections before calling `ShutdownAll`)
7. The new worker's health check either connects to the wrong (old) worker (bearer token mismatch → fail) or the socket never appears reliably
8. After 10s timeout: `"worker 0 failed to start: timeout waiting for worker socket"` → `"too many connections"`

This only triggers when:
- `max_workers` is set (enables the worker semaphore that gates concurrency)
- Connections arrive during the brief transition window where old and new workers coexist

## Fix

Add a **per-CP-instance socket prefix** using the process PID. Worker sockets are now named `w{pid}-{id}.sock` instead of `worker-{id}.sock` (e.g., `w12345-0.sock`). Each CP instance's workers use unique paths, so old and new workers never collide.

**Changed:** `controlplane/worker_mgr.go` — added `socketPrefix` field set to `w{pid}` in `NewFlightWorkerPool`, used in `SpawnWorker` for socket path generation.

## How to reproduce

The new test `TestHandoverWithMaxWorkersRapidConnections` reproduces this reliably. Without the fix:

```bash
# Revert the fix to see the failure:
# git stash && go test -v -timeout 120s -count=1 ./tests/controlplane/ -run TestHandoverWithMaxWorkersRapidConnections

# The test:
# 1. Starts a CP with max_workers=3
# 2. Warms up a connection (spawns a worker)
# 3. Sends SIGUSR1 to trigger handover
# 4. Immediately fires 20 connections staggered 100ms apart — hitting the transition window
# 5. Verifies all 20 succeed, then verifies 5 more after settling
#
# Before the fix: 1+ connections fail with "pq: too many connections"
# because SpawnWorker(0) on the new CP races with old CP's worker-0.sock
#
# After the fix: 20/20 succeed because new CP uses w{pid}-0.sock, no collision
```

Run all handover + max_workers tests:
```bash
go test -v -timeout 120s -count=1 ./tests/controlplane/ -run "TestHandoverWithMaxWorkers"
```

Full control plane suite (14 tests):
```bash
go test -v -timeout 300s -count=1 ./tests/controlplane/
```

## Test plan

- [x] `TestHandoverWithMaxWorkers` — basic handover with max_workers set
- [x] `TestHandoverWithMaxWorkersAndActiveConns` — handover with saturated workers
- [x] `TestHandoverWithMaxWorkersRapidConnections` — 20 rapid connections during transition (the repro case)
- [x] All 11 existing controlplane tests still pass
- [x] Ran rapid connections test 3 additional times to confirm no flakiness

### DIFF
--- a/controlplane/worker_mgr.go
+++ b/controlplane/worker_mgr.go
@@ -39,6 +39,7 @@ type FlightWorkerPool struct {
 	workers      map[int]*ManagedWorker
 	nextWorkerID int // auto-incrementing worker ID
 	socketDir    string
+	socketPrefix string // unique per CP instance to avoid collisions during handover
 	configPath   string
 	binaryPath   string
 	maxWorkers   int           // 0 = unlimited
@@ -52,12 +53,13 @@ type FlightWorkerPool struct {
 func NewFlightWorkerPool(socketDir, configPath string, maxWorkers int) *FlightWorkerPool {
 	binaryPath, _ := os.Executable()
 	pool := &FlightWorkerPool{
-		workers:    make(map[int]*ManagedWorker),
-		socketDir:  socketDir,
-		configPath: configPath,
-		binaryPath: binaryPath,
-		maxWorkers: maxWorkers,
-		shutdownCh: make(chan struct{}),
+		workers:      make(map[int]*ManagedWorker),
+		socketDir:    socketDir,
+		socketPrefix: fmt.Sprintf("w%d", os.Getpid()),
+		configPath:   configPath,
+		binaryPath:   binaryPath,
+		maxWorkers:   maxWorkers,
+		shutdownCh:   make(chan struct{}),
 	}
 	if maxWorkers > 0 {
 		pool.workerSem = make(chan struct{}, maxWorkers)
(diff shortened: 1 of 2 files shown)


EDsCODE:
Closing — the root cause (worker socket path collision during handover) was inadvertently fixed by #239 (pre-bind worker sockets in CP process). The pre-bind approach eliminates the race because the new CP atomically removes + rebinds the socket before the worker starts, so old and new workers never contend on the same path. The test from this PR would still be useful as a regression test but the fix itself is no longer needed.

PR #410, opened 2026-04-10
Fall back to no-LIMIT schema probe when LIMIT 0 triggers DuckDB bugs

benben:
## Summary

\`read_parquet('https://...large_file.parquet')\` fails with \`Invalid Error: stoi\` in control-plane mode but works in standalone mode. Root cause: DuckDB's stock v1.5.1 httpfs extension calls \`std::stoi\` on HTTP \`Content-Length\` headers without error handling (\`httpfs_curl_client.cpp:254\`). The \`LIMIT 0\` code path in \`GetQuerySchema\` triggers different HTTP request patterns in DuckDB's parquet reader that hit this bug, while the same query without \`LIMIT 0\` works fine.

Standalone mode never calls \`GetQuerySchema\` — the simple query protocol sends queries directly to DuckDB. Control-plane mode uses Flight SQL, which always probes the schema first via \`GetQuerySchema\` → \`LIMIT 0\`.

## Fix

When the \`LIMIT 0\` probe fails with an error containing "stoi" or "stoll", retry with the original query (no LIMIT). \`database/sql\`'s \`Rows.Close()\` cancels the in-progress scan after reading column types, so no full table scan occurs.

## Verified live on EKS

Built locally with Apple \`container\`, pushed to \`ghcr.io/benben/duckgres:fix-httpfs-stoi-76536e5\`, patched into the \`duckgres\` namespace deployment.

Worker log showing the fallback:
\`\`\`
level=WARN msg="LIMIT 0 schema probe failed, retrying without LIMIT."
  error="Invalid Error: stoi"
  query_prefix="SELECT count(*) FROM read_parquet('https://datasets.clickhouse.com/hits_compatible/hits.parquet'..."
\`\`\`

Query result:
\`\`\`
 count_star()
--------------
     99997497
(1 row)
\`\`\`

The full 99,997,497-row ClickBench hits dataset loaded successfully over HTTPS through the control-plane Flight SQL path.

## Test plan

- [x] \`go test ./duckdbservice/...\` — all green
- [x] \`go vet ./duckdbservice/...\` — clean
- [x] Live test on EKS: \`read_parquet('https://datasets.clickhouse.com/hits_compatible/hits.parquet', binary_as_string=True)\` returns 99,997,497 rows
- [x] Fallback log line confirmed in worker pod logs
- [x] Worker stays alive throughout (health check fix from #409 also deployed)

### DIFF
--- a/duckdbservice/arrow_helpers.go
+++ b/duckdbservice/arrow_helpers.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"fmt"
+	"log/slog"
 	"math/big"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


benben:
fixed the extension instead, working on getting this compiled and shipped to duckgres

PR #426, opened 2026-04-17
chore: Map DuckDB error prefixes to specific PostgreSQL SQLSTATEs

abhischekt:
Every DuckDB query error previously surfaced as 42000, forcing PG-aware clients (psycopg2, SQLAlchemy, dlt, ORMs) into the generic SyntaxErrorOrAccessRuleViolation bucket even for missing tables, missing columns, unique violations, etc. classifyErrorCode now branches on DuckDB's exception-type prefix and maps CATALOG/BINDER/PARSER/CONVERSION/ OUTOFRANGE/CONSTRAINT/PERMISSION/TRANSACTION/DEPENDENCY errors to the appropriate SQLSTATE (42P01, 3F000, 42703, 42883, 22P02, 23505, ...).

The DuckDB Go driver does not expose a typed ExceptionType, so the classifier parses the error string — this is the same approach as the existing isDropTableOnViewError helper.

Add unit coverage for every branch and a wire-level test that drives canonical DuckDB errors through handleQuery and asserts the SQLSTATE on the ErrorResponse frame.

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -366,7 +366,9 @@ func isDuckLakeMetadataConnectionLost(err error) bool {
 
 // classifyErrorCode returns the most appropriate PostgreSQL SQLSTATE for a
 // DuckDB error. Transaction conflicts get 40001 (serialization_failure), which
-// signals PG-aware clients to retry. Query cancellations get 57014.
+// signals PG-aware clients to retry. Query cancellations get 57014. Remaining
+// errors are classified by DuckDB's exception-type prefix; the prefix is the
+// only signal the Go driver exposes today, so we parse the message string.
 func classifyErrorCode(err error) string {
 	if isQueryCancelled(err) {
 		return "57014"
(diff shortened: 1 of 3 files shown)


abhischekt:
FYI I worked around this with a monkeypatch in `dlt`:
```
def _patch_dlt_exception_mapper_for_duckgres() -> None:
    """Teach dlt's Postgres client to recognise Duckgres "missing relation" errors.

    Duckgres proxies SQL through Flight/Arrow and reports both missing schemas
    and missing tables as ``psycopg2.errors.SyntaxErrorOrAccessRuleViolation``
    with a ``Catalog Error: ... does not exist`` payload, rather than the
    ``UndefinedTable``/``InvalidSchemaName`` codes that dlt's exception mapper
    special-cases as "first run, nothing here yet". Without this rewrap, every
    bootstrap-time introspection (``_dlt_pipeline_state``, ``_dlt_version``)
    surfaces as a terminal pipeline failure on a fresh dataset.

    The patch is idempotent and falls through to the original mapper for any
    error it can't classify, so vanilla Postgres pipelines remain unaffected.
    """
    if getattr(Psycopg2SqlClient, "_duckgres_patch_installed", False):
        return

    original_mapper = Psycopg2SqlClient._make_database_exception

    @classmethod
    def _duckgres_aware_mapper(cls, ex: Exception) -> Exception:
        if "does not exist" in str(ex):
            return DatabaseUndefinedRelation(ex)
        return original_mapper(ex)

    Psycopg2SqlClient._make_database_exception = _duckgres_aware_mapper
    Psycopg2SqlClient._duckgres_patch_installed = True
```

benben (approved):
:shipit:

bill-ph (changes requested):
Four issues stood out in adversarial review:

1. `server/conn.go`: schema-qualified missing-table errors are misclassified. DuckDB currently emits `Catalog Error: Table with name "missing_schema.t" does not exist because schema "missing_schema" does not exist.`, but `catalogErrorCode` checks `table with name` before the schema-specific case, so this returns `42P01` instead of `3F000`.

2. `server/conn.go`: `binderErrorCode` still misses common undefined-table/function cases. Live DuckDB emits `Binder Error: Referenced table "t" not found!` for bad aliases and `Binder Error: No function matches ...` for overload mismatches; both currently fall through to `42601`. The new tests only cover missing/ambiguous columns, and the overload case in `server/transient_test.go` is modeled under `Catalog Error`, which does not match current DuckDB output.

3. `server/conn.go`: every `Conversion Error:` is forced to `22P02`, but DuckDB also uses that prefix for numeric overflow casts. For example, `SELECT CAST(1000 AS TINYINT)` yields a conversion error with an out-of-range condition, so this path reports `invalid_text_representation` instead of `22003`.

4. `server/conn.go`: the transaction-state mapping keys off `Transaction Error:`, but DuckDB currently emits `TransactionContext Error:` for common cases like `BEGIN; BEGIN;`. Those still fall through to `42000`, so the PR misses the main transaction-state errors clients will actually see.

I validated the examples above against live `duckdb` output while reviewing the PR branch locally in a disposable worktree.

bill-ph (approved):

PR #452, opened 2026-04-23
controlplane: reconcile stranded worker pods + fix shutdown ordering

benben:
## Summary

Two prod workers (IDs 31758, 31763) from deploy \`198b09d\` — predating the patched-httpfs pin at #411 — survived two subsequent rollouts with DB rows already at state=\`retired\`, still serving queries with the stoi-overflow-prone vanilla httpfs. Clients that landed on them hit \"Invalid Error: stoi\" on >2 GB parquet reads; the rest of prod is on the patched extension.

Root cause is a gap between \`ShutdownAll\` and the janitor's orphan sweep:

1. \`ShutdownAll\` marked each worker \`retired\` in the DB **before** issuing the K8s pod delete, and the delete was fire-and-forget (\`_ = ...Delete\`). On API hiccup or CP SIGKILL mid-shutdown, the pod survived while the DB row was already terminal.
2. \`ListOrphanedWorkers\` excludes \`retired\`/\`lost\` rows (they'd loop the janitor on 404s), and bare worker pods carry no owner reference, so nothing else in the cluster reaps them.

\`configstore/store.go:847\` already claimed this leak \"is handled authoritatively by the K8s label-based startup scan in K8sWorkerPool.cleanupOrphanedWorkerPods\". That function did not exist in the codebase.

## Changes

**Fix 1 — \`ShutdownAll\` 3-step CAS chain per worker**

\`\`\`
MarkWorkerDraining (idle/hot_idle/... → draining)
  → pods.Delete()
  → RetireDrainingWorker (draining → retired)
\`\`\`

- Intermediate \`draining\` state fences the worker against claims from other CPs (their claim queries match \`state=idle\`/\`hot_idle\`), eliminating the mid-delete claim race.
- On pod-delete failure the row stays in \`draining\`. \`ListOrphanedWorkers\` already includes \`draining\` in its cleanup states, so orphan cleanup retires+deletes the pod automatically once the CP's heartbeat expires (~50s).
- \`NotFound\` on delete is treated as success.
- On \`MarkWorkerDraining\` CAS miss (row already terminal or owned by another CP) we skip the worker entirely.
- Split \`markWorkerRetiredLocked\` into the original (persists via Upsert) plus a new \`markWorkerRetiredInMemoryLocked\` (lifecycle + metrics only) so the CAS-chain state isn't clobbered by an unconditional upsert.

**Fix 2 — \`cleanupOrphanedWorkerPods\` reconciler**

- Lists worker pods by \`app=duckgres-worker\` label every janitor tick (leader-only).
- Deletes pods whose DB row is \`retired\`/\`lost\` or missing, with a 2-minute age grace so newly-spawned pods (which create the pod before upserting the row) aren't racy-deleted.
- Idempotent: \`NotFound\` counts as success; next tick retries anything missed.
- Catches the two stranded prod pods on the first janitor tick after deploy — no \`kubectl delete\`, no CP rollout gymnastics.

**Belt-and-suspenders, not overlap:** Fix 1 keeps the happy shutdown path clean (no pods get stranded in the first place). Fix 2 handles every other failure mode (kubelet hangs after accepting Delete, CP OOMKilled before reaching the final CAS, historical stranded pods from before this fix).

## Test plan

All tests pass under \`go test -tags kubernetes ./controlplane/...\` locally.

- [x] 7 reconciler cases: retired/lost/missing rows delete pod; idle row left alone; young pod skipped (age grace); NotFound counts as success; non-worker pods untouched.
- [x] 5 \`ShutdownAll\` cases: happy 3-step chain; strict \`draining→delete→retired\` ordering; CAS-miss skips the worker entirely; pod-delete failure leaves row in \`draining\`; NotFound counts as success.
- [x] 2 janitor cases: reconciler is invoked every tick; runs after the version reaper.
- [x] \`go vet -tags kubernetes\` clean.
- [x] Non-Kubernetes build (\`go build ./controlplane/...\`) still compiles.

## Manual verification after deploy

- The two stale pods \`duckgres-5bfc46d74f-worker-31758\` and \`-31763\` in \`posthog-mw-prod-us/duckgres\` should be deleted within ~5s of a janitor leader with this build.

### DIFF
--- a/controlplane/configstore/store.go
+++ b/controlplane/configstore/store.go
@@ -666,6 +666,58 @@ func (cs *ConfigStore) RetireIdleWorker(workerID int, reason string) (bool, erro
 	return result.RowsAffected > 0, nil
 }
 
+// MarkWorkerDraining atomically transitions a worker into the draining state
+// if and only if it is still owned by the caller and not already terminal. It
+// returns true when the transition happened.
+//
+// Used by ShutdownAll to fence a worker before issuing its K8s pod delete: no
+// other CP can claim the worker once it's draining (ClaimIdleWorker and
+// ClaimHotIdleWorker filter on state=idle and state=hot_idle respectively),
+// so the pod-delete/DB-retire chain can proceed without a claim race. If the
+// CP then crashes before the final retired transition, ListOrphanedWorkers
+// includes draining rows whose owner CP has expired, so orphan cleanup
+// retires the worker and deletes the pod.
+//
+// The ownerCPInstanceID guard prevents a stale CP from moving a worker that
+// has already been taken over by a successor.
+func (cs *ConfigStore) MarkWorkerDraining(workerID int, ownerCPInstanceID string) (bool, error) {
+	drainableStates := []WorkerState{
+		WorkerStateSpawning,
+		WorkerStateIdle,
+		WorkerStateReserved,
+		WorkerStateActivating,
+		WorkerStateHot,
+		WorkerStateHotIdle,
+	}
+	result := cs.db.Table(cs.runtimeTable((&WorkerRecord{}).TableName())).
+		Where("worker_id = ? AND owner_cp_instance_id = ? AND state IN ?", workerID, ownerCPInstanceID, drainableStates).
+		Updates(map[string]any{
+			"state":      WorkerStateDraining,
+			"updated_at": time.Now(),
+		})
+	if result.Error != nil {
+		return false, fmt.Errorf("mark worker %d draining: %w", workerID, result.Error)
+	}
+	return result.RowsAffected > 0, nil
+}
+
+// RetireDrainingWorker atomically transitions a draining worker to retired.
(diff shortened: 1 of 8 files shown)


bill-ph:
Review finding:

`cleanupOrphanedWorkerPods` assumes a missing runtime row comes back as `rec == nil`, but the real `ConfigStore.GetWorkerRecord` wraps `gorm.ErrRecordNotFound` and returns it as an error. In production that means this path logs and skips instead of deleting the pod, so the new "missing DB row" recovery path is unreachable outside the fake test store.

The current test passes because `captureRuntimeWorkerStore.GetWorkerRecord` returns `(nil, nil)` for absent rows:
- `controlplane/k8s_pool.go:313-316`
- `controlplane/configstore/store.go:509-514`
- `controlplane/k8s_pool_test.go:2268-2280`

I didn’t find another concrete issue beyond this one.

bill-ph (approved):

PR #455, opened 2026-04-23
feat(controlplane): per-warehouse PgBouncer opt-in for Duckling CR

fuziontech:
## Summary

Adds `pgbouncer.enabled` to the `ManagedWarehouse` config and plumbs it through the provisioning controller so operators can turn on the per-Duckling PgBouncer pooler (provisioned by the Crossplane composition in [PostHog/charts#10400](https://github.com/PostHog/charts/pull/10400)) for a **specific org** without a global flag or a manual `kubectl patch`.

Complements [#448](https://github.com/PostHog/duckgres/pull/448), which handles the *reader* side (workers routing through `status.metadataStore.pgbouncerEndpoint`). This PR is the *writer* side — the control plane sets `spec.metadataStore.pgbouncer.enabled: true` on the Duckling CR when an org is opted in.

### Data model
- `ManagedWarehousePgBouncer{Enabled bool}` — new type in `configstore/models.go`.
- Embedded on `ManagedWarehouse` with GORM prefix `pgbouncer_`, yielding column `pgbouncer_enabled`. `AutoMigrate` handles the new column on startup; existing rows default to `false`.
- Mirrored in `ManagedWarehouseConfig` + `copyManagedWarehouseConfig`.

### Admin API
- `managedWarehouseRequest` accepts a new top-level `pgbouncer: {enabled: bool}` JSON field on `PUT /managed-warehouses/:id`.
- `managedWarehouseUpsertColumns()` allowlist updated so the upsert actually persists the flag.

Example:
```bash
curl -X PUT -H 'Authorization: Bearer $ADMIN' \
  https://duckgres.example/admin/managed-warehouses/<org-id> \
  -d '{"pgbouncer":{"enabled":true}}'
```

### Provisioner
- `DucklingClient.Create(ctx, orgID, MinACU, MaxACU)` refactored to `Create(ctx, orgID, CreateOptions{...})` — room to grow without more positional args.
- New field `CreateOptions.PgBouncerEnabled`; when true, the generated CR carries `spec.metadataStore.pgbouncer.enabled: true`.
- Controller reads `w.PgBouncer.Enabled` and passes it through. Log line at CR-create surfaces the flag for audit.

### Scope

This PR only affects **Create** time — new Ducklings respect the flag. Existing CRs are **not** reconciled by the control plane (no `Update` method on `DucklingClient`). For existing orgs mid-flight, flip the flag in the DB (or via admin API), then either:
- recreate the Duckling, or
- `kubectl -n ducklings patch duckling <name> --type=merge -p '{"spec":{"metadataStore":{"pgbouncer":{"enabled":true}}}}'` manually (the CP won't revert the manual edit).

A spec-reconcile follow-up can land separately if we decide we want full declarative drift correction.

## Test plan

- [x] `go build -tags kubernetes ./...` — clean (one pre-existing unrelated failure in `tests/perf`, confirmed on `main`).
- [x] `go vet -tags kubernetes ./controlplane/...` — clean.
- [x] `go test -tags kubernetes ./controlplane/...` — all green (admin-suite Postgres-container test skipped locally due to missing `docker-compose`, same pre-existing note as #448).
- [x] `TestReconcilePendingCreatesCR` — unchanged path, now also asserts no `pgbouncer` block when disabled.
- [x] `TestReconcilePendingEmitsPgBouncerBlock` — new, asserts `spec.metadataStore.pgbouncer.enabled=true` appears on the CR when flag set.
- [ ] Manual canary after charts#10400 and this PR land: flip the PostHog-org's `pgbouncer_enabled`, verify CR regenerates with the block, confirm duckgres#448's worker picks up the endpoint.

## Rollback

Revert this PR. Zero orgs have `pgbouncer_enabled=true` by default, so reverting is a no-op. Individual orgs can be opted out by flipping the flag back to `false` and recreating the Duckling.

### DIFF
--- a/controlplane/admin/api.go
+++ b/controlplane/admin/api.go
@@ -306,6 +306,7 @@ func managedWarehouseUpsertColumns() []string {
 		"metadata_store_port",
 		"metadata_store_database_name",
 		"metadata_store_username",
+		"pgbouncer_enabled",
 		"s3_provider",
 		"s3_region",
 		"s3_bucket",
@@ -402,6 +403,7 @@ type apiHandler struct {
 type managedWarehouseRequest struct {
 	WarehouseDatabase              configstore.ManagedWarehouseDatabase          `json:"warehouse_database"`
 	MetadataStore                  configstore.ManagedWarehouseMetadataStore     `json:"metadata_store"`
+	PgBouncer                      configstore.ManagedWarehousePgBouncer         `json:"pgbouncer"`
 	S3                             configstore.ManagedWarehouseS3                `json:"s3"`
 	WorkerIdentity                 configstore.ManagedWarehouseWorkerIdentity    `json:"worker_identity"`
 	WarehouseDatabaseCredentials   configstore.SecretRef                         `json:"warehouse_database_credentials"`
@@ -428,6 +430,7 @@ func (r managedWarehouseRequest) toManagedWarehouse() configstore.ManagedWarehou
 	return configstore.ManagedWarehouse{
 		WarehouseDatabase:              r.WarehouseDatabase,
 		MetadataStore:                  r.MetadataStore,
+		PgBouncer:                      r.PgBouncer,
 		S3:                             r.S3,
 		WorkerIdentity:                 r.WorkerIdentity,
 		WarehouseDatabaseCredentials:   r.WarehouseDatabaseCredentials,
--- a/controlplane/admin/api.go
+++ b/controlplane/admin/api.go
@@ -3,8 +3,10 @@
 package admin
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
+	"io"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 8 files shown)


fuziontech:
Fixed in `aa6c01b`.

The admin handler now preserves existing managed-warehouse fields when a request only updates `pgbouncer`, instead of decoding into a zero-valued struct and overwriting the full row. In practice, `{"pgbouncer":{"enabled":true}}` now behaves like a partial update for existing warehouses and no longer clears metadata store, S3, secret refs, or provisioning state.

Also added a regression test covering the toggle-only path.

Validation:
- `go test -tags kubernetes ./controlplane/admin -run '^(TestPutWarehouse.*|TestGetWarehouseReturnsNotFoundWhenMissing|TestGetOrgIncludesWarehouse|TestListOrgsIncludesWarehouse)$' -count=1 -timeout 60s`
- full admin package test run is still blocked by the existing `docker-compose` requirement in `TestUpsertManagedWarehousePreservesCreatedAt`

PR #574, opened 2026-05-19
feat(provisioner): Lakekeeper-as-Iceberg-catalog — PR1 (provisioning machinery)

fuziontech:
## Summary

First of three PRs adding Lakekeeper as the default Iceberg catalog for duckgres workers. This PR is provisioning scaffolding only — no worker behavior change.

PR1 (this PR) builds:
- A thin Lakekeeper REST client (`Info`, `Bootstrap`, `EnsureWarehouse`)
- A Postgres admin helper (`EnsureDatabase`) for creating the per-org `lakekeeper_<orgid>` database inside the org's existing managed-warehouse Aurora cluster
- K8s `EnsureCR` / `EnsureSecret` / `GetCR` helpers for the `Lakekeeper` CR from the lakekeeper-operator
- The `LakekeeperProvisioner.EnsureForOrg` reconciler that chains it all together idempotently
- Schema additions to `ManagedWarehouseIceberg` (Backend selector + Lakekeeper-side config fields)
- A default-deny baseline NetworkPolicy in the `lakekeeper` namespace

PR2 (next) will add the worker-side ATTACH SQL and activation-payload wiring. PR3 will swap allowall + NetworkPolicy isolation for OIDC SA-token auth.

## Background

Prototype validation at `tmp/lakekeeper-proto/` (not in this PR) empirically confirmed:
- DuckDB 1.5.2 + the bundled iceberg extension can `ATTACH` a Lakekeeper REST catalog and CREATE/INSERT/SELECT through it end-to-end (parquet/manifest/snapshot all land in S3).
- Lakekeeper warehouse storage profile needs `sts-enabled: true`, `flavor: s3-compat` (for MinIO; `aws` for AWS), and `remote-signing-enabled: false` for DuckDB to authenticate to S3.
- Token rotation in DuckDB-iceberg is handled internally via the OAuth2 client_credentials flow — caller only manages long-lived CLIENT_ID/CLIENT_SECRET.

A `FINDINGS.md` writeup of the prototype is intentionally kept in `tmp/` so it's clear it's not the source of truth for production code.

## Design decisions captured

- **Default for new orgs**: `ManagedWarehouseIceberg.Backend` defaults to `"lakekeeper"`. Legacy `"s3_tables"` stays as an explicit escape hatch. `ResolvedBackend()` handles the empty-string case for rows that predate the column.
- **Reuse org's existing Aurora**: per-org `lakekeeper_<orgid>` database, not a dedicated PG cluster. The provisioner takes the admin DSN as an injected dependency so the secret-lookup logic isn't baked in here — PR2 wires it to the K8s Secret managed by Crossplane.
- **Lazy provisioning**: triggered from the worker activator at first activation. The reconciler is synchronous and idempotent; `ErrBootstrapPending` signals "requeue, not fail".
- **Status preservation in EnsureCR**: spec drift correction never clobbers operator-set status. The fix was load-bearing for the fake-client tests but is also the right contract for real K8s.
- **CAS-mismatch is benign**: configstore gained a typed `ErrWarehouseStateMismatch` sentinel. EnsureForOrg returns `nil` on a CAS miss because the Lakekeeper REST write is already committed by then.

## Test plan

- [ ] CI passes both default and `-tags kubernetes` test suites
- [ ] Manual `EnsureForOrg` against a dev cluster with the lakekeeper-operator installed (PostHog/charts companion PR)
- [ ] Live integration tests against the prototype stack at `tmp/lakekeeper-proto/` (`PG_ADMIN_DSN`, `LAKEKEEPER_SMOKE_URL` env-gated) pass locally
- [ ] No worker behavior change — ducklings still use the S3 Tables path on every existing org

## Companion PR

PostHog/charts: `feat(lakekeeper-operator): vendor chart + add dev AppSet` — installs the operator + CRD on `posthog-dev`. The duckgres provisioner assumes the CRD is available; the charts PR provides it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ data-test/
 bench-results/
 /cache-proxy
 /cmd/cache-proxy/cache-proxy
+
+# Local prototype/scratch dirs (e.g. tmp/lakekeeper-proto/)
+tmp/
--- a/controlplane/configstore/iceberg_backend_test.go
+++ b/controlplane/configstore/iceberg_backend_test.go
@@ -1,0 +1,23 @@
+package configstore
+
+import "testing"
+
+func TestManagedWarehouseIceberg_ResolvedBackend(t *testing.T) {
+	cases := []struct {
+		name  string
+		in    ManagedWarehouseIceberg
+		want  string
+	}{
+		{"empty defaults to lakekeeper", ManagedWarehouseIceberg{}, IcebergBackendLakekeeper},
+		{"explicit lakekeeper", ManagedWarehouseIceberg{Backend: IcebergBackendLakekeeper}, IcebergBackendLakekeeper},
+		{"explicit s3_tables", ManagedWarehouseIceberg{Backend: IcebergBackendS3Tables}, IcebergBackendS3Tables},
+		{"unknown passthrough", ManagedWarehouseIceberg{Backend: "future"}, "future"},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if got := c.in.ResolvedBackend(); got != c.want {
+				t.Errorf("ResolvedBackend() = %q, want %q", got, c.want)
+			}
+		})
+	}
+}
--- a/controlplane/configstore/models.go
+++ b/controlplane/configstore/models.go
@@ -150,9 +150,20 @@ type ManagedWarehouseIceberg struct {
 
 	// Lakekeeper fields (Backend == "lakekeeper"). Populated by the
 	// provisioner after the per-org Lakekeeper is ready.
-	LakekeeperEndpoint        string `gorm:"size:512" json:"lakekeeper_endpoint,omitempty"`
+	LakekeeperEndpoint string `gorm:"size:512" json:"lakekeeper_endpoint,omitempty"`
+
+	// LakekeeperWarehouse is the warehouse NAME (e.g. "org-acme"), not the
+	// UUID. Iceberg REST clients pass this as the `warehouse` parameter to
+	// /v1/config and the server returns the UUID as a prefix for subsequent
+	// calls. PR2's worker-side ATTACH SQL uses this value directly.
 	LakekeeperWarehouse       string `gorm:"size:128" json:"lakekeeper_warehouse,omitempty"`
 	LakekeeperClientID        string `gorm:"size:128" json:"lakekeeper_client_id,omitempty"`
+
+	// LakekeeperOAuth2ServerURI is the OAuth2 token endpoint URI for the
+	// duckling-side CREATE SECRET. Empty during PR1 (allowall mode);
+	// populated by PR3 once OIDC SA-token auth is wired. PR2 worker code
+	// must guard against empty and either skip the OAuth2 fields on the
+	// CREATE SECRET statement or emit a different secret shape.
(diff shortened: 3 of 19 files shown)


fuziontech:
## orbstack e2e validation — three production bugs caught

Stood the full pipeline up against a real orbstack k8s cluster + real lakekeeper-operator + real Postgres + real MinIO. Three bugs surfaced that the dynamic-fake clientset and httptest mocks couldn't catch (all fixed in 3d26bca):

1. **Postgres 15+ public-schema permissions** — the freshly-created `lakekeeper_<orgid>` database's `public` schema is owned by the admin user, and PG 15 removed default CREATE perms for non-owner roles. Lakekeeper's `migrate` step CrashLoopBackOffs forever because it can't CREATE TABLE. `EnsureRole` now does `ALTER DATABASE OWNER TO <role>` + `ALTER SCHEMA public OWNER TO <role>` inside the target database via a re-DSN helper.

2. **TLS mismatch on PG connection** — the operator's `sslMode` defaults to `require`, but local Postgres has no TLS. Added `PGSSLMode` to both `LakekeeperCRSpec` and `ProvisioningInputs`. Production leaves it empty and inherits the secure default; local/dev passes `disable`.

3. **`/management/v1/*` is at the server root, not under `/catalog`** — the provisioner was passing `baseURL + "/catalog"` to the management HTTP client, assembling `<x>/catalog/management/v1/warehouse` which 404s. `/catalog` is only for the Iceberg REST API that ducklings hit later. Now the management client uses the bare baseURL; the persisted `LakekeeperEndpoint` still includes `/catalog` for DuckDB.

## E2E result

```
=== RUN   TestE2E_LakekeeperProvisionerOnOrbstack
    orgID=e2e1779223717
    [0s] bootstrap pending, requeueing...
    [3s] bootstrap pending, requeueing...
    ...
    [18s] bootstrap pending, requeueing...
    EnsureForOrg succeeded after 21.5s
    warehouse row: endpoint=http://lakekeeper-e2e1779223717.lakekeeper.svc:8181/catalog
                   warehouse=org-e2e1779223717
                   client_id=duckling-e2e1779223717
--- PASS: TestE2E_LakekeeperProvisionerOnOrbstack (21.74s)
```

Full pipeline reconciles in ~20s including image pull. Reproducible across runs. The e2e test is gated by `LAKEKEEPER_E2E_KUBECONFIG` so CI won't run it without a cluster, but it's a one-command verification path for anyone bringing up the local stack:

```
LAKEKEEPER_E2E_KUBECONFIG=\$HOME/.kube/config \
PG_ADMIN_DSN='postgres://lakekeeper:lakekeeper@localhost:5434/lakekeeper?sslmode=disable' \
  go test -tags kubernetes ./controlplane/provisioner/ -run TestE2E -v
```

The Lakekeeper image is pulled from `quay.io/lakekeeper/catalog:latest` (works); the **operator** image at `quay.io/lakekeeper/lakekeeper-operator:v0.0.1` doesn't exist upstream — the chart we vendored pins a tag that was never published. Local validation worked by building the operator from the source we have (`docker build -t quay.io/lakekeeper/lakekeeper-operator:v0.0.1-local .` inside `lakekeeper-operator/`). The charts PR (#PostHog/charts/11164) will need to either un-pin the operator image or coordinate an upstream release before this can deploy.

PR #578, opened 2026-05-19
Adjust warm targets after capacity misses

bill-ph:
## Summary
- trigger async warm backfill after warm-capacity misses, including runtime-store durable-claim misses
- layer temporary demand targets on top of chart/env warm-pool targets
- coalesce retries while a warm backfill is in flight so retry loops do not stampede pod creation or ratchet targets on every attempt

## Tests
- go test -count=1 ./controlplane
- go test -count=1 -tags kubernetes ./controlplane
- go test -count=1 ./tests/k8s -run TestIsTransientDBError

### DIFF
--- a/controlplane/control.go
+++ b/controlplane/control.go
@@ -675,7 +675,7 @@ func sessionCreationErrorResponse(err error) (code string, message string) {
 			retryAfter = DefaultWarmCapacityRetryAfter
 		}
 		retrySeconds := int((retryAfter + time.Second - 1) / time.Second)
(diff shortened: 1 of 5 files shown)


bill-ph:
Closing this as superseded by the revised dynamic warm-capacity plan. The current PR ties foreground warm-capacity misses directly to local target bumps / async backfill, with local in-flight state as the controller. After review, we want the production path to be staged differently: keep #577 as the cold-reject baseline, add typed miss reasons, record capacity misses centrally, and let the janitor/reconciler compute a bounded decaying warm target before spawning. That avoids retry-driven pod creation and lets multi-CP demand aggregate correctly.

PR #624, opened 2026-05-26
Fix truncation of identifiers longer than 63 bytes in the transpiler

fuziontech:
## Problem

After upgrading 0.4 → 1.0, queries against tables whose names exceed **63 bytes** fail with errors like:

```
Catalog Error: Table with name reports__cohort_user_period_activity_calendar_monthly__41192770 does not exist!
Did you mean "reports__cohort_user_period_activity_calendar_monthly__4119277097"?
```

Note the requested name is exactly **63 chars** while the real table is **65** — a classic PostgreSQL `NAMEDATALEN` truncation.

## Root cause

The 0.4 → 1.0 upgrade replaced regex-based query rewriting with an **AST-based transpiler** (`fe9d446`). The transpiler parses SQL with `pg_query_go` → libpg_query, which is PostgreSQL's own C scanner and enforces `NAMEDATALEN` (63 bytes), silently truncating any longer identifier during `Parse`. DuckDB has no such limit, so a `Parse → Deparse` round trip renames the table to its 63-char prefix and the lookup fails.

This only triggers for queries that take the transpiler's full parse path (i.e. that contain a PostgreSQL-specific feature such as a `::regtype` cast, a pg function, `public.`, etc.). Plain queries bypass the parser and were unaffected — which is why it looked size/shape dependent.

## Fix

`transpiler/longident.go` swaps any identifier longer than 63 bytes for a short, unique placeholder **before** parsing, then restores the originals on the SQL emitted from the AST. The scan is SQL-aware — it skips string literals, E-strings, dollar-quoted bodies, comments, and `$N` parameters, and handles both bare and `"double-quoted"` identifiers. For normal short queries it is a no-op (early return).

Wired into:
- `transpiler/transpiler.go` `transpileWithFlags` — the core path used by every `Transpile()` caller, including the control-plane Flight path, prepared statements, and writable-CTE / SQL-override rewrites.
- `server/conn.go` `handleMultiStatementQuery` — a separate standalone-mode deparse that also truncated.

## Tests

- New `transpiler/longident_test.go` covers the production case, bare/quoted/qualified identifiers, dedup, multiple distinct names, placeholder-prefix collision avoidance, and that string/dollar/comment/E-string contents are **not** mis-protected.
- End-to-end (validated locally against a real DuckDB, not committed to keep the transpiler package driver-free): create the 65-char table → transpile a query that forces the parse path → DuckDB finds it. Fails on `main`, passes here.
- Full `transpiler/...` and `server/` suites pass; `gofmt`/`go vet` clean.

## Known residual (out of scope)

A server-side cursor whose **inner** SELECT references a >63-char table still truncates, because it deparses from a tree parsed upstream. It's a narrow edge case; can be a follow-up if needed.

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -1347,7 +1347,7 @@ func (c *clientConn) handleQuery(body []byte) error {
 	// Each statement gets its own results, with a single ReadyForQuery at the end.
 	tree, parseErr := pg_query.Parse(query)
 	if parseErr == nil && len(tree.Stmts) > 1 {
-		return c.handleMultiStatementQuery(tree)
+		return c.handleMultiStatementQuery(query)
 	}
 
 	// Transpile PostgreSQL SQL to DuckDB-compatible SQL
@@ -1847,7 +1847,19 @@ func (c *clientConn) executeSelectQuery(query string, cmdType string) (int64, st
 // each statement gets its own RowDescription/DataRow/CommandComplete messages,
 // with a single ReadyForQuery at the end. If any statement fails, remaining
 // statements are skipped.
-func (c *clientConn) handleMultiStatementQuery(tree *pg_query.ParseResult) error {
+func (c *clientConn) handleMultiStatementQuery(query string) error {
+	// Re-parse with long identifiers protected so that splitting the batch via
+	// Deparse below does not silently truncate names > 63 bytes (see
+	// transpiler/longident.go). executeSingleStatement transpiles each
+	// statement again, which re-applies the same protection.
+	parseSQL, longIdents := transpiler.ProtectLongIdentifiers(query)
+	tree, err := pg_query.Parse(parseSQL)
+	if err != nil {
+		c.sendError("ERROR", "42601", fmt.Sprintf("syntax error: %v", err))
+		_ = wire.WriteReadyForQuery(c.writer, c.txStatus)
+		_ = c.writer.Flush()
+		return nil
+	}
 	slog.Debug("Multi-statement simple query.", "user", c.username, "count", len(tree.Stmts))
 
 	for _, stmt := range tree.Stmts {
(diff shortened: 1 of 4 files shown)


fuziontech:
## Self-review (multi-angle)

Ran a multi-angle review. One real bug found and fixed in `ce9257b`:

**Boundary-aware restoration (fixed).** The original `restoreLongIdentifiers` used `strings.ReplaceAll`, which also rewrote a placeholder *embedded* in a larger token. The writable-CTE transform builds a temp-table name from the CTE name (`_cte_<name>_<hex>`), so a **quoted, >63-byte CTE name** caused restore to splice the original quoted text into the middle of the generated name → invalid SQL:

```
CREATE TEMP TABLE "_cte_"My CTE With Spaces..."_ec35953f" AS (...)   -- broken
```

Fix: only replace placeholder tokens that stand alone (not flanked by identifier chars). Real references restore as before; placeholders embedded in generated names are left intact (valid and self-consistent across CREATE/SELECT/DROP). Added a regression test plus an adversarial round-trip/no-panic test.

### Considered and accepted as-is
- **Cursor inner queries** (`DECLARE … CURSOR FOR SELECT … <long table>`) still truncate — they deparse from a tree parsed before protection. Already called out as a known residual in the PR description; not on the production (DML) path. Can be a follow-up.
- **Bare mixed-case >63-byte names** restore verbatim rather than lowercased. DuckDB matches unquoted names case-insensitively, so this resolves correctly; it's strictly better than the previous truncation.
- **Quoted→bare placeholder in restrictive grammar positions**: worst case is a parse failure → `FallbackToNative`, which sends the original SQL (with the real long name) to DuckDB, which has no length limit. Safe by construction.
- **Multibyte boundary**: protection triggers on >63 *bytes*, matching libpg_query's truncation point; over-protection is a harmless no-op round trip.

PR #630, opened 2026-05-28
Bump DuckDB 1.5.2 → 1.5.3 (driver + bundled extensions)

fuziontech:
## Summary

Bumps the DuckDB engine across the standalone server, control plane, and worker images from **1.5.2 → 1.5.3**, moving the Go driver/bindings, the bundled extensions (httpfs, ducklake), and the CI matrix in lock-step. Also switches the httpfs source from `benben/duckdb-httpfs` to `PostHog/duckdb-httpfs` (the PostHog fork is now the canonical home, with a tag-driven release pipeline).

## Why now / what's in 1.5.3

DuckDB 1.5.3 is a feature/bugfix release in the 1.5 line — **no file-format changes, no breaking C-API changes** for our embedding. Notable bits that touch us:

- **Iceberg extension** gains `MERGE INTO`, `INSERT`/`UPDATE` on partitioned tables, `ALTER TABLE`, `GEOMETRY` support, and new schema-property functions — directly relevant to the in-flight Lakekeeper/ducklings work.
- **AWS extension** gains `web_identity` chain support (relevant to per-org Lakekeeper EKS Pod Identity).
- **jemalloc** is now statically linked into core on Linux (was a separate extension); transparent for us.
- **DuckLake** stays on 1.0.
- **Quack protocol** ships as a new core extension that auto-installs on first use — we don't touch it, but flagging since auto-install would attempt a network fetch.

## What changed (all in lock-step on the same DuckDB version)

| Pin | Before | After |
|---|---|---|
| `go.mod` duckdb-go/v2 + bindings + 5 platform libs | `v2.10502.0` / `v0.10502.0` | `v2.10503.0` / `v0.10503.0` |
| `Dockerfile{,worker}` `DUCKDB_EXTENSION_VERSION` | `1.5.2` | `1.5.3` |
| `Dockerfile{,worker}` `HTTPFS_EXTENSION_TAG` | `v1.5.2-stoi-fix` | `v1.5.3-stoi-fix` |
| httpfs source URL | `benben/duckdb-httpfs` | **`PostHog/duckdb-httpfs`** |
| `Dockerfile{,worker}` `DUCKLAKE_EXTENSION_TAG` | `v1.0-posthog.2` | `v1.0-posthog.3` |
| `container-image-worker-cd.yml` default matrix row | `1.5.2` | `1.5.3` (1.5.2 demoted to fallback; 1.5.1 dropped) |
| `scripts/ducklake_version_matrix.sh` `DEFAULT_VERSIONS` | `v2.10502.0 v2.10501.0` | `v2.10503.0 v2.10502.0` |
| `tests/k8s/k8s_test.go::expectedDucklakeExtensionVersion` | `4a9d2f00` | `e4ac5150` (short SHA of `PostHog/ducklake@v1.0-posthog.3`) |
| `server/bundled_extensions_test.go` paths | `v1.5.2` | `v1.5.3` |
| Inline comments in `server/server.go` + `server/profiling_test.go` | — | bumped |

## Pre-req releases (both cut today)

- ✅ **`PostHog/duckdb-httpfs` v1.5.3-stoi-fix** — release published with `httpfs-linux-{amd64,arm64}.duckdb_extension` (https://github.com/PostHog/duckdb-httpfs/releases/tag/v1.5.3-stoi-fix). Cut with `skip_tests: true` because two upstream tests are brittle/environmental, not defects in the Content-Length patch:
  - `test_disabled_file_system_httpfs.test` — harness double-registers the `memory` secret storage in the static-load + reload path; also needs live `github.com`.
  - `glob_s3_recursive.test_slow` — asserts an exact HTTP request count (got 5, expected 4), shifted by 1.5.3's curl/httpfs request pattern. Data is read correctly.

  The binary compiled cleanly and **5414/5416 assertions still passed**; the real exercise is via this repo's `tests/integration` + `tests/k8s`.

- ⏳ **`PostHog/ducklake` v1.0-posthog.3** — tag pushed, build+publish run is in progress at the time of opening this PR (https://github.com/PostHog/ducklake/actions). **CI on this PR will be red until that release lands** (the docker build `curl`s the release asset). Re-run CI once the ducklake release exists.

## Verified locally
- `go build ./...` clean on 1.5.3 (CGO ABI compatible).
- `go test -c -tags 'k8s_integration kubernetes' ./tests/k8s/` builds — TestMain in that package is destructive against the active kubeconfig (see `tests/k8s/CLAUDE.md`), so I'm not running the suite from here. Full k8s validation belongs in `just test-k8s-integration` against a kind cluster.

## Risks / things to keep an eye on
- Real proof of the httpfs binary loading correctly under duckgres is in the k8s integration suite once it runs.
- If `expectedDucklakeExtensionVersion = "e4ac5150"` doesn't match what `duckdb_extensions().extension_version` returns from the newly-built ducklake (in case the build embeds a different format), the k8s ducklake-fork-verification test will fail with a clear message naming the actual value — easy to update.
- Memory profile may shift slightly with jemalloc now statically linked vs as an extension — worth a glance at worker RSS post-deploy, but not expected to be material.


.github/workflows/container-image-worker-cd.yml
```diff
@@ -36,15 +36,15 @@ jobs:
             fail-fast: false
             matrix:
                 duckdb:
+                    - version: "1.5.3"
+                      go: "v2.10503.0"
+                      bindings: "v0.10503.0"
+                      httpfs: "v1.5.3-stoi-fix"
+                      default: true
                     - version: "1.5.2"
```
bill-ph, line 81: is this building duckdb 1.5.2 with ducklake 1.5.3?

fuziontech: Good catch — yes, that was a real bug. The build-args block passed `go`, `bindings`, `httpfs`, and DuckDB version per-row, but `DUCKLAKE_EXTENSION_TAG` was hard-coded as a Dockerfile default, so both rows were pulling the same ducklake (`v1.0-posthog.4`, built against DuckDB 1.5.3). The 1.5.2 fallback would have ended up with DuckDB 1.5.2 + a 1.5.3-compiled ducklake — an ABI mismatch that the docker build itself wouldn't catch (it only verifies the Go bindings pin; extension files are just downloaded), so it would have surfaced only at `LOAD ducklake` runtime.

Fixed in d63977a: added a `ducklake` field to each matrix row and passed it through as a build-arg. 1.5.3 stays on `v1.0-posthog.4`; 1.5.2 fallback now pins to `v1.0-posthog.2` (which was compiled for 1.5.2). New CD run dispatched: https://github.com/PostHog/duckgres/actions/runs/26580536815

PR #640, opened 2026-05-29
fix(controlplane): default passthrough sessions into the tenant catalog

fuziontech:
## Problem

Portola runs sqlmesh against the MTCP through a **passthrough** user. Their job fails with `Catalog "portola" does not exist`, and more subtly, passthrough sessions land in the wrong catalog entirely.

Verified live against prod-us (`entirely-chief-wildcat.dw.us.postwh.com`, user `duckgres-duckdb`):

```
6/6 fresh passthrough connections:  current_database() = memory
ducklake.core.membership_events  → 7,170,968 rows   ✅ (data is here)
```

On the remote-worker path, passthrough sessions skip the entire session-init block (`controlplane/control.go`, gated `if !passthroughUser`), so the worker's DuckDB connection is left in DuckDB's default **empty in-memory `memory` catalog**. Consequences:

- `current_database()` returns `memory` — so sqlmesh's default-catalog auto-detection (which probes `current_catalog`) resolves to the wrong, ephemeral catalog.
- Unqualified DDL/DML executes against `memory` and **never reaches the tenant's DuckLake warehouse**.

Standalone passthrough doesn't have this problem — `server.CreatePassthroughDBConnection` calls `setDuckLakeDefault` (`USE ducklake`) / `setIcebergDefault`. The remote-worker path simply had no equivalent.

## Fix

Add a passthrough branch to the session-init block that detects the attached `ducklake` catalog and issues the matching `USE`, preferring an iceberg-default user's configured catalog (mirrors the standalone behavior):

```go
passthroughSessionDefaultCatalogCommand(defaultCatalog, duckLakeAttached):
  defaultCatalog == "iceberg"  -> USE iceberg.public
  duckLakeAttached             -> USE ducklake
  otherwise                    -> "" (leave session as-is)
```

After this, a passthrough session reliably reports `current_database() = ducklake` and lands in the tenant warehouse.

### Deliberately *not* in scope

This does **not** enable logical-catalog mapping for passthrough. Those sessions still talk raw DuckDB against the **physical** catalog name (`ducklake`); `SetLogicalCatalogMapping` stays disabled for them (guarded with `&& !passthroughUser`). Making the logical org name (e.g. `portola`) resolve for passthrough would require re-introducing transpiler/rewrite behavior, which defeats the purpose of passthrough and can't reliably handle the DuckDB-native SQL these users send. The customer-side counterpart to this fix is to reference the real catalog name `ducklake` in their sqlmesh project (duckdb dialect), while keeping the PG connection database as their org name for routing.

## Testing

- `go build -tags kubernetes ./controlplane/...`
- `go vet -tags kubernetes ./controlplane/`
- `go test -tags kubernetes ./controlplane/` (incl. new `TestPassthroughSessionDefaultCatalogCommand`)

### Verify post-deploy
Connect to MTCP as a passthrough user and confirm `SELECT current_database()` returns the tenant catalog (`ducklake`) on a fresh session, and that an unqualified `CREATE TABLE` lands in the warehouse rather than `memory`.

### DIFF
--- a/controlplane/control.go
+++ b/controlplane/control.go
@@ -1126,9 +1126,12 @@ func (cp *ControlPlane) handleConnection(conn net.Conn) {
 		}
 	}()
 
-	// Passthrough users skip pg_catalog initialization and DuckLake catalog
-	// detection — they bypass the PG compatibility layer entirely, so the
-	// metadata setup that drives logical catalog mapping is unused for them.
+	// Passthrough users skip pg_catalog initialization and logical-catalog
+	// mapping — they bypass the PG compatibility layer entirely. They still
+	// need a default catalog, though: without one the worker session stays in
+	// DuckDB's empty in-memory catalog, so current_database() reports "memory"
+	// and unqualified DDL/DML never reaches the warehouse (see the passthrough
+	// branch below).
 	var duckLakeAttached bool
 	if !passthroughUser {
 		initCtx, initCancel := context.WithTimeout(context.Background(), cp.cfg.SessionInitTimeout)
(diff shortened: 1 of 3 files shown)


fuziontech:
## Review pass (self) + fixes

Independent review found **no correctness blockers**. Verified: `initCtx`/`initCancel` handled on every path; the shared `err` is nil on the passthrough success path and not read after the block; `SetLogicalCatalogMapping(cc, duckLakeAttached && !passthroughUser)` preserves the exact pre-PR behavior (flag stays false for passthrough); iceberg-before-ducklake ordering is intentional; bare `USE ducklake` is correct for passthrough (no memory.main macros to preserve, matching standalone `setDuckLakeDefault`).

Applied two should-fix items (commit 477e6e9):
- Replaced the repeated `"ducklake"` literals with a `physicalDuckLakeCatalog` constant.
- Passthrough no longer silently discards a client connect-time `search_path` — it logs a warning (passthrough clients can `SET search_path` themselves).

**Deferred (fast-follow, not blocking):** a `tests/k8s/` assertion that a passthrough session reports `current_database()=ducklake` and resolves the warehouse through the real activation pipeline. This change is confined to the CP connection handler (not the activation payload / attach paths that CLAUDE.md flags), existing k8s ducklake tests remain valid, and the helper logic is unit-tested. I'll verify the behavior directly against prod-us after promotion.

PR #678, opened 2026-06-04
Fix file persistence session initialization

### LINKED ISSUE #561: file_persistence: true breaks standalone mode — "No catalog + schema named memory found"
When enabling `file_persistence: true` in standalone mode, every connection fails immediately after authentication
   with:

  `Catalog Error: SET schema: No catalog + schema named "memory" found.`

  The file-backed DuckDB database is opened successfully (`Opening file-backed DuckDB.
  path="/data/duckgres/test.duckdb?allow_unsigned_extensions=true"`), extensions load, but the session setup
  crashes when trying to switch to the `memory` catalog.

  ## Steps to Reproduce

  1. Configure `duckgres.yaml`:
  ```yaml
  host: "0.0.0.0"
  port: 5432
  data_dir: "/data/duckgres"
  file_persistence: true

  tls:
    cert: "/certs/server.crt"
    key: "/certs/server.key"

  extensions:
    - parquet
    - httpfs

  users:
    test: "test"
 ```
  2. Run duckgres in standalone mode (default)
  3. Connect with psql:
  `psql "host=localhost user=test dbname=test sslmode=require"`

### Test Setup
  - Image: ghcr.io/posthog/duckgres:latest
  - Version: build-c6f9f04756b68cd4100b65374f2dff1b0f37657c
  - Mode: standalone (default)
  - DuckDB extensions: parquet, httpfs
  - Duckgres is running in a swarm cluster

sueun-dev:
This fixes standalone file-backed connections failing during session metadata initialization when the `memory` catalog is not attached yet.

What changed:
- Attach an in-memory `memory` catalog on the missing-memory-catalog path before installing session metadata views.
- Restore standalone file-persistence sessions to the persisted physical catalog/search path after metadata setup.
- Add regression coverage for metadata initialization and a PG-wire reconnect path with `file_persistence` enabled.

Why:
Issue #561 reports `file_persistence: true` failing during connection startup with `No catalog + schema named "memory" found`. The metadata setup needs the compatibility views in `memory`, but standalone file-backed connections can start without that catalog attached.

Checked:
- `go test -run TestInitSessionDatabaseMetadataWorksOnFileBackedDatabase ./server/sessionmeta`
- `go test -run TestFilePersistenceStandaloneConnectionInitializesSession ./server`
- `go test ./server/sessionmeta ./server`
- `just build`
- `just lint`
- `just test-unit`
- Not run: Docker-backed integration tests because Docker is not installed in this local environment.

Fixes #561


benben:
@sueun-dev thanks for contributing! Can you please take a look at these changes so we get this into a mergeable shape?

- Rebase: conflicts with main (conn.go split, sessionmeta.go changed)
- Pass real catalog (c.username), not c.database: views filter on current_database(), so user tables invisible in pg_tables/information_schema.tables, fixes the silently-failing restore-defer too and makes the conn.go hunk
  unnecessary
- Test: assert created table visible in information_schema.tables
- Codex comment is valid and should be addressed

sueun-dev:
Reworked along your notes — thanks for the concrete list.

- Rebased onto main (merged it in; conn.go and sessionmeta.go both moved).
- `InitSessionDatabaseMetadata` now receives the file-backed catalog, so its own deferred restore leaves the session there and the extra `USE`/`search_path` block in conn.go is gone. Gated on `DataDir != ""`, since an empty DataDir makes `DuckDBDSN` fall back to `:memory:` where there is no catalog named after the user — that was the codex comment, and it was right.
- The session test asserts a freshly created table is visible in `information_schema.tables`, and that `current_database()` is the backing catalog rather than the client's dbname. Reverting the catalog selection makes it fail with `current_database() = "test", want "testuser"`, so it pins the actual symptom.

`go test ./server/...` is green.

PR #740, opened 2026-06-10
feat(controlplane): configurable hot-idle TTL + wire dead DUCKGRES_K8S_WORKER_MAX_TTL env

fuziontech:
## Why

Part of the defensive follow-ups to the sqlmesh worker-bootstrap stall incident (see #737, #738, #739).

`defaultHotIdleTTL` is a hardcoded 5-minute const (`controlplane/multitenant.go`). Default-profile workers (the big exclusive 46cpu/360Gi ones — which have no per-worker `ttl_minutes`) therefore retire 5m after going idle, so a tenant running **hourly** jobs cold-starts every run: pod spawn + tenant activation on every connection burst. The client-side `duckgres.worker_ttl` GUC is **not** a workaround for these tenants: setting any `duckgres.*` option demotes the session to a sized profile clamped to 16cpu/64Gi (`allowClientExclusiveNode=false` in prod), shrinking their workers.

## What

- **`DUCKGRES_K8S_HOT_IDLE_TTL`** (env-only, like other K8s knobs): overrides the janitor's hot-idle retention for workers without a per-worker TTL. Unset/0 = existing 5m default — behavior is byte-identical without the env var. Sized-profile workers (own `ttl_minutes`) are unaffected, per the existing `CASE WHEN ttl_minutes > 0` in `ListExpiredHotIdleWorkers`.
- **Wires `DUCKGRES_K8S_WORKER_MAX_TTL`**: the duckgres Helm chart emits this env var for the `worker_ttl` clamp (`K8sConfig.WorkerMaxTTL`), but no Go code ever consumed it — the chart value was dead config and the clamp silently stayed unbounded. Now resolved in `configresolve` and mapped in both binaries (all-in-one `main.go` + `cmd/duckgres-controlplane`).

## Tests

- `configresolve`: parse both env vars; invalid/negative values warn and resolve to 0 (default).
- `controlplane`: `effectiveHotIdleTTL` fallback semantics (0 → 5m, configured wins, negative → 5m).
- `go test ./configresolve/... && go test -tags kubernetes ./controlplane/` pass; `go vet` + `gofmt` clean on touched files.

## e2e harness

Per the CLAUDE.md escape hatch: with the env unset (how mw-dev deploys), behavior is byte-identical to today, and asserting a changed TTL in-Job would require ≥TTL of wall-clock waiting to observe janitor retire timing — not practical for the per-PR harness. The existing hot-idle reclaim assertions continue to exercise the (default) TTL path.

## Follow-up

Chart PR to add a `hotIdleTtl` value → `DUCKGRES_K8S_HOT_IDLE_TTL` env in the duckgres chart, and an mw-prod-us value (e.g. `70m`) once we choose the cost trade-off (idle r6gd nodes held longer per org).


fuziontech:
Scope clarification after reviewing the post-#710 architecture (warm pool removed; sized profiles carry their own per-worker TTL): this knob only governs **default-profile** workers — connections that send no sizing GUCs, i.e. external tenants with plain connection strings. Sized/colocated workers are unaffected (their TTL comes from `duckgres.worker_ttl`, default 20m, clamped by the WorkerMaxTTL this PR wires up).

For tenant-level warm retention the preferred lever is the per-org default worker profile (PR incoming on `claude/org-default-worker-profile`), which lets an org default to a small sized worker with its own TTL — cheap to hold warm. This knob remains useful as the fleet-wide fallback for whatever stays on the default exclusive shape.

PR #752, opened 2026-06-10
feat: per-user persistent secret manager (CREATE PERSISTENT SECRET) for MTCP

fuziontech:
## Summary

A customer can now run `CREATE PERSISTENT SECRET` and have it survive across sessions and worker pods in the multitenant remote backend. Worker pods are ephemeral (pinned `secret_directory` wiped on recycle), so until now any client-issued secret died with the session.

```sql
CREATE PERSISTENT SECRET my_gcs (TYPE gcs, KEY_ID '…', SECRET '…', SCOPE 'gs://my-bucket');
-- reconnect later, different worker pod:
SELECT name FROM duckdb_secrets();   -- my_gcs is back
DROP PERSISTENT SECRET my_gcs;       -- durably gone
```

Plain / `TEMPORARY` `CREATE SECRET` stays session-scoped passthrough, unchanged.

## Design

- **Intercept at the CP** (`server/conn_user_secrets.go`, classifier in `server/usersecrets/`): both simple and extended protocols, passthrough users included. Execute on the live session first (DuckDB validates), **then** persist — a store failure after a successful exec returns an explicit "applied but will NOT survive this session" error. `DROP` deletes the store row even when the session-side drop fails, so a broken stored secret can't become unremovable.
- **Store in the config store**: new `duckgres_org_user_secrets` table keyed `(org_id, username, secret_name)`; the statement is AES-256-GCM-encrypted with the row key bound as AAD (ciphertext can't be moved between rows). Direct DB reads (not snapshot-polled), so a secret set via one CP replica is immediately live on all.
- **Replay at session creation**: the user's decrypted statements ride the existing TLS+bearer `CreateSession` worker RPC. The worker first **wipes all persistent-storage secrets** (system secrets `ducklake_s3`/`iceberg_sigv4`/`iceberg_oauth` are in-memory and untouched), then replays. The wipe is the cross-user isolation step for hot-idle worker reuse and failing it fails the session; replay failures degrade to warnings so a stale secret can't lock a user out.
- **Key management**: env-only `DUCKGRES_USER_SECRET_KEY` (base64 32-byte AES key, mounted from a K8s Secret). Unset → `CREATE PERSISTENT SECRET` fails with a clear 0A000 ("not enabled on this deployment"); malformed → CP refuses to start; `DROP` works keyless so stale rows stay removable after key loss.
- **Guardrails**: reserved system names, unnamed persistent secrets, >8KB statements, >32 secrets/user are all rejected before anything executes.
- **Log hygiene**: `usersecrets.RedactForLog` now guards `logQueryStarted/Finished/Error`, the query log, trace spans, and `pg_stat_activity` — secret DDL appears as `CREATE PERSISTENT SECRET my_s3 (…redacted)`. This also closes the pre-existing leak where a passthrough `CREATE SECRET` landed in the query log with plaintext credentials.

## Test plan

- Unit: classifier/crypto/redaction (`server/usersecrets`), CP interception semantics (`server/conn_user_secrets_test.go`), worker wipe+replay against real DuckDB (`duckdbservice/user_secrets_test.go`), configstore CRUD against the Postgres test container.
- e2e (`tests/e2e-mw-dev/harness.sh`): `persistent_user_secret` on **both** cnpg and ext lanes (create → cross-session replay → reserved/unnamed rejected → durable drop) and `persistent_user_secret_isolation` on cnpg (a second org user must not see root's secret on a reused hot-idle worker; root's stored copy survives the other user's session). `run.sh` mints a per-run key into `duckgres-tokens` and the CP env.
- `go build` / `go vet` / `go test ./...` clean on both plain and `-tags kubernetes` builds. (Two pre-existing failures unrelated to this PR: `TestCatalogPsqlCommands/psql_dn` fails identically on main; `controlplane/admin` postgres-container test needs `docker-compose` on PATH.)

## Deployment note

Real deployments need a **stable** `DUCKGRES_USER_SECRET_KEY` minted and mounted before this is live (rotating it orphans stored user secrets — they're skipped with a loud log and removable via `DROP`). Until the key ships, the feature is cleanly disabled. Chart/infra PRs follow.

### DIFF
--- a/controlplane/session_mgr.go
+++ b/controlplane/session_mgr.go
@@ -61,6 +61,12 @@ type SessionManager struct {
 	activeSlots    int
 	waiters        []*connectionWaiter
 	limiter        connectionLimiter
+
+	// userSecretLoader returns the user's persistent CREATE SECRET statements
+	// (decrypted) to replay on a worker at session creation. nil outside the
+	// multitenant/remote backend. A loader error must not block the session:
+	// callers log and continue without secrets.
+	userSecretLoader func(ctx context.Context, username string) ([]string, error)
 }
 
 type flightReconnectPool interface {
@@ -92,6 +98,12 @@ func (sm *SessionManager) SetMaxConnections(n int) {
 	defer sm.mu.Unlock()
 	sm.maxConnections = n
 	sm.grantWaitersLocked()
+}
+
+// SetUserSecretLoader installs the per-user persistent-secret loader used at
+// session creation (multitenant/remote backend only).
+func (sm *SessionManager) SetUserSecretLoader(loader func(ctx context.Context, username string) ([]string, error)) {
+	sm.userSecretLoader = loader
 }
 
 // SetConnectionLimiter replaces the local process limiter with a cluster-wide
@@ -407,7 +419,21 @@ func (sm *SessionManager) createSessionOnWorker(ctx context.Context, username st
 		"owner_cp_instance_id", worker.OwnerCPInstanceID(),
 		"owner_epoch", worker.OwnerEpoch(),
 	)
-	sessionToken, err := worker.CreateSession(ctx, username, memoryLimit, threads)
+	// Load the user's persistent secrets for replay. Failure to load degrades
+	// to a session without user secrets (logged loudly) rather than a refused
+	// connection: the config store being briefly unavailable must not lock
+	// every returning user out of their warehouse.
+	var secretStatements []string
+	if sm.userSecretLoader != nil {
+		var secretErr error
+		secretStatements, secretErr = sm.userSecretLoader(ctx, username)
+		if secretErr != nil {
+			slog.Error("Failed to load user persistent secrets; session starts without them.",
+				"pid", pid, "worker", worker.ID, "user", username, "error", secretErr)
+		}
+	}
+
+	sessionToken, secretWarnings, err := worker.CreateSession(ctx, username, memoryLimit, threads, secretStatements)
 	if err != nil {
 		slog.Warn("Failed to create session on worker.",
 			"pid", pid,
@@ -422,6 +448,11 @@ func (sm *SessionManager) createSessionOnWorker(ctx context.Context, username st
 			sm.pool.RetireWorkerIfNoSessions(worker.ID)
 		}
 		return 0, nil, fmt.Errorf("create session on worker %d: %w", worker.ID, err)
+	}
+
+	for _, w := range secretWarnings {
+		slog.Warn("User persistent secret replay warning.",
+			"pid", pid, "worker", worker.ID, "user", username, "warning", w)
 	}
 
 	executor := flightclient.NewFlightExecutorFromClient(worker.client, sessionToken)
--- a/duckdbservice/service.go
+++ b/duckdbservice/service.go
@@ -911,8 +911,9 @@ func (svc *DuckDBService) WaitForDrain(ctx context.Context) bool {
 
 // CreateSession creates a new DuckDB session for the given username.
 // secretStatements are the user's persistent secrets to replay (shared-warm
-// mode only); replay failures come back as client-facing warnings, not
-// errors.
+// mode only); replay failures come back as warnings — logged on the worker
+// and again by the control plane's session manager, never failing the
+// session — rather than errors.
 func (p *SessionPool) CreateSession(username, memoryLimit string, threads int, secretStatements []string) (*Session, []string, error) {
 	start := time.Now()
 	finishDrain, drainErr := p.beginDrainWork(false)
@@ -1036,7 +1037,20 @@ func (p *SessionPool) CreateSession(username, memoryLimit string, threads int, s
 	// plane. Replay failures degrade to warnings; a wipe failure fails the
 	// session because handing user A's secrets to user B is not acceptable.
 	var secretWarnings []string
-	if p.sharedWarmMode {
+	switch {
+	case p.sharedWarmMode && p.maxSessions != 1:
+		// The wipe and replay below are only safe because exactly one session
+		// runs at a time (DuckDB secrets are instance-global; k8s workers are
+		// spawned with DUCKGRES_DUCKDB_MAX_SESSIONS=1). Assert that here, at
+		// the point of reliance, instead of trusting a flag set two layers
+		// away: with any other cap, wiping would delete a concurrent
+		// session's secrets mid-query. Skip hygiene entirely and scream.
+		slog.Error("User persistent secret hygiene requires max_sessions=1 on shared-warm workers; skipping wipe+replay.",
+			"max_sessions", p.maxSessions, "user", username, "secrets", len(secretStatements))
+		if len(secretStatements) > 0 {
+			secretWarnings = []string{"persistent secrets were NOT restored: worker session cap is misconfigured (requires max_sessions=1)"}
+		}
+	case p.sharedWarmMode:
 		secretCtx, secretCancel := context.WithTimeout(context.Background(), userSecretOpTimeout)
 		wiped, wipeErr := wipeUserPersistentSecrets(secretCtx, conn)
 		if wipeErr != nil {
--- a/duckdbservice/service.go
+++ b/duckdbservice/service.go
@@ -910,11 +910,14 @@ func (svc *DuckDBService) WaitForDrain(ctx context.Context) bool {
 }
 
 // CreateSession creates a new DuckDB session for the given username.
-func (p *SessionPool) CreateSession(username, memoryLimit string, threads int) (*Session, error) {
+// secretStatements are the user's persistent secrets to replay (shared-warm
+// mode only); replay failures come back as client-facing warnings, not
+// errors.
+func (p *SessionPool) CreateSession(username, memoryLimit string, threads int, secretStatements []string) (*Session, []string, error) {
 	start := time.Now()
 	finishDrain, drainErr := p.beginDrainWork(false)
 	if drainErr != nil {
-		return nil, drainErr
+		return nil, nil, drainErr
 	}
 	defer finishDrain()
 
@@ -922,7 +925,7 @@ func (p *SessionPool) CreateSession(username, memoryLimit string, threads int) (
 	p.mu.Lock()
 	if p.maxSessions > 0 && len(p.sessions)+p.reserved >= p.maxSessions {
 		p.mu.Unlock()
-		return nil, fmt.Errorf("max sessions reached (%d)", p.maxSessions)
+		return nil, nil, fmt.Errorf("max sessions reached (%d)", p.maxSessions)
 	}
 	p.reserved++
 	p.mu.Unlock()
@@ -958,7 +961,7 @@ func (p *SessionPool) CreateSession(username, memoryLimit string, threads int) (
 			p.mu.Lock()
 			p.reserved--
 			p.mu.Unlock()
-			return nil, cfgErr
+			return nil, nil, cfgErr
 		}
 
 		// Fallback: create a shared DB if warmup failed or wasn't run.
@@ -979,7 +982,7 @@ func (p *SessionPool) CreateSession(username, memoryLimit string, threads int) (
 			p.mu.Lock()
 			p.reserved--
 			p.mu.Unlock()
-			return nil, fmt.Errorf("create session db: %w", p.fallbackErr)
+			return nil, nil, fmt.Errorf("create session db: %w", p.fallbackErr)
 		}
 		db = p.fallbackDB
 		slog.Info("Using fallback DB (warmup failed).", "duration", time.Since(fallbackStart))
@@ -997,7 +1000,7 @@ func (p *SessionPool) CreateSession(username, memoryLimit string, threads int) (
 		p.mu.Lock()
 		p.reserved--
 		p.mu.Unlock()
-		return nil, fmt.Errorf("failed to obtain connection from pool (timeout after 30s): %w", err)
+		return nil, nil, fmt.Errorf("failed to obtain connection from pool (timeout after 30s): %w", err)
 	}
 	slog.Debug("Acquired DB connection from pool.", "user", username, "duration", time.Since(start))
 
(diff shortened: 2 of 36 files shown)


EDsCODE (approved):

server/conn_user_secrets.go
```diff
@@ -0,0 +1,184 @@
...
+		c.logQueryFinished(redacted, queryStart, 0, queryFinalErr)
+	}()
+
+	upperQuery := strings.ToUpper(query)
+	cmdType := c.getCommandType(upperQuery)
+
+	_, execErr := c.executor.ExecContext(ctx, query)
+	if execErr != nil {
+		queryFinalErr = execErr
+		if st.Kind == usersecrets.KindDrop {
```
fuziontech: **[review:correctness — high] DROP fallback swallows every exec error class, not just "replay never applied it".**

This branch fires on ANY `execErr`: user cancellation, a Flight RPC blip, an aborted-transaction error (`25P02`), worker draining, or DuckDB's temp/persistent ambiguity error ("secret exists in multiple storages"). In all of those the store row is deleted and the client gets a success tag while the live session-side secret still exists and keeps working until session end.

Worst case is a credential revocation: `DROP SECRET leaked_key` races a transient RPC error → client gets positive confirmation of a revocation that didn't happen; the live worker keeps using the revoked secret for the rest of the session. A cancelled or aborted-txn DROP also irreversibly deletes the persisted row with no rollback.

Suggest matching only the error this fallback exists for — DuckDB's secret-not-found message (same pattern as `isWorkerSessionCapError`) — and surfacing every other exec error unchanged.

server/conn_user_secrets.go
```diff
@@ -0,0 +1,184 @@
...
+		return true, "", &secretDDLError{classifyErrorCode(execErr), errMsg}
+	}
+
+	// Session-side DDL succeeded; now make it durable. Failures here must be
+	// surfaced as errors — the client believing a secret persisted when it
+	// didn't is the one outcome this feature exists to prevent.
+	var storeErr error
+	switch st.Kind {
+	case usersecrets.KindCreate:
+		storeErr = mgr.PutSecret(ctx, c.orgID, c.username, st.Name, query)
```
fuziontech: **[review:correctness] `CREATE PERSISTENT SECRET IF NOT EXISTS` overwrites the store even when DuckDB no-ops.**

When the secret already exists, DuckDB returns success without changing the live secret — but `PutSecret` unconditionally replaces the stored statement. The current session keeps the old credentials while every future session replays the new ones: the `IF NOT EXISTS` guard the user relied on is silently inverted into a replace at the persistence layer.

Suggest skipping `PutSecret` when the statement has `IF NOT EXISTS` and the store already holds that name (or rejecting `IF NOT EXISTS` for persistent secrets outright with a clear error).

server/usersecrets/redact.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	// Only the statement head matters for matching; don't uppercase a
+	// potentially huge query just to reject it. 64 bytes comfortably covers
+	// the longest prefix plus the word-boundary byte.
+	probe := head
+	if len(probe) > 64 {
+		probe = probe[:64]
+	}
+	upper := strings.ToUpper(probe)
+	for _, prefix := range createSecretPrefixes {
```
fuziontech: **[review:security — high] Flat prefix match is looser than Classify's tokenizer → redaction can miss statements that ARE intercepted.**

`Classify` tolerates arbitrary whitespace/newlines/comments between keywords (tokenizer), but this match requires the exact single-space uppercase prefix. So `CREATE  PERSISTENT SECRET s (…)` (double space, newline, or an inline comment between keywords — e.g. from a SQL formatter) is intercepted and persisted, yet `RedactForLog` returns it verbatim: full credentials land in INFO-level lifecycle logs, the query log, the OTel span, and `pg_stat_activity`. A redaction false negative is exactly what this function's own comment says must never happen.

Suggest driving redaction off the same tokenizer (e.g. `Classify(query).Kind == KindCreate` → redact from the first `(`), so the redactor can never be out of sync with the interceptor. That also removes the second hand-maintained prefix list (already drifted: this one has `CREATE TEMP SECRET`, `isDuckDBUtilityCommand`'s does not).

server/conn_extended_query.go
```diff
@@ -546,7 +551,15 @@ func (c *clientConn) handleExecute(body []byte) {
 	cmdType := c.getCommandType(upperQuery)
 	returnsResults := queryReturnsResults(p.stmt.query)
 
-	slog.Debug("Execute portal.", "user", c.username, "portal", portalName, "params", len(args), "query", p.stmt.query)
+	// Intercept persistent-secret DDL (multitenant remote backend): persist /
+	// delete the user's stored secret alongside the session-side DDL. Uses
+	// the original (untranspiled) text — secret DDL is DuckDB-native and
+	// always falls back unmodified. ReadyForQuery is sent by Sync.
+	if c.handleUserSecretDDLExtended(p.stmt.query) {
```
fuziontech, line 558: **[review:correctness] Bound portal parameters are discarded for intercepted secret DDL.**

`Classify` never inspects the option list, so `CREATE PERSISTENT SECRET s (TYPE s3, KEY_ID $1, SECRET $2)` with bound params is intercepted here, and `execUserSecretDDL` runs `ExecContext(ctx, query)` with no args — the pre-PR path passed `args...`. Parameterizing credential values is exactly what a security-conscious driver does. And if a parameterized form ever executed successfully, the persisted statement text would contain literal `$1` placeholders that fail on every future replay.

Suggest explicitly rejecting parameterized persistent-secret DDL with a clear error ("literal values required so the statement can be replayed"), rather than executing with unbound params.

duckdbservice/service.go
```diff
@@ -1025,6 +1028,32 @@ func (p *SessionPool) CreateSession(username, memoryLimit string, threads int) (
...
+	// different user of the same org would otherwise see the previous user's
+	// persistent secrets. Wipe first (mandatory — this is the cross-user
+	// isolation step), then replay this user's secrets from the control
+	// plane. Replay failures degrade to warnings; a wipe failure fails the
+	// session because handing user A's secrets to user B is not acceptable.
+	var secretWarnings []string
+	if p.sharedWarmMode {
+		secretCtx, secretCancel := context.WithTimeout(context.Background(), userSecretOpTimeout)
+		wiped, wipeErr := wipeUserPersistentSecrets(secretCtx, conn)
+		if wipeErr != nil {
```
fuziontech, line 1056: **[review:robustness] Wipe failure refuses the session and retires the worker — fine — but the one-session precondition the wipe depends on is never asserted.**

Two different gates guard the same invariant: the create-time wipe runs on `sharedWarmMode` alone, the destroy-time wipe adds `&& p.maxSessions == 1`. The wipe (and replay) are only safe because k8s workers ship `DUCKGRES_DUCKDB_MAX_SESSIONS=1` — enforced two layers away in `k8s_pool.go`. If anyone ever raises that cap, session B's create-time wipe deletes session A's replayed secrets mid-query, with no error at the point of misconfiguration.

Suggest asserting it here: refuse (or loudly log) wipe+replay when `sharedWarmMode && p.maxSessions != 1`, so the assumption is enforced where it's relied on.

Also worth noting from the verifier pass: a hot-idle worker holding a secret whose `DROP PERSISTENT SECRET IF EXISTS` repeatably errors will fail the reuse session, get retired, respawn, replay the same secret, and repeat — bounded churn rather than a hard loop (fresh workers wipe trivially), but worth a metric on wipe failures.

controlplane/session_mgr.go
```diff
@@ -424,6 +450,11 @@ func (sm *SessionManager) createSessionOnWorker(ctx context.Context, username st
 		return 0, nil, fmt.Errorf("create session on worker %d: %w", worker.ID, err)
 	}
 
+	for _, w := range secretWarnings {
```
fuziontech, line 453: **[review:UX/docs] Replay warnings never reach the client, but two doc comments claim they do.**

`SessionPool.CreateSession`'s comment says replay failures come back as "client-facing warnings", and the warnings are plumbed through four layers (pool → flight response → `ManagedWorker.CreateSession` → here) — then only `slog.Warn`'d, duplicating the worker-side log. The user whose secret failed to replay (missing extension, key-rotation skip in `LoadStatements`, config-store blip in the loader above) gets zero indication; their first S3 query fails with a baffling 403 and nothing ties it to the lost secret.

Either finish the job (emit a `NoticeResponse` on session start — the plumbing is already in place, which is the hard part) or drop the four-layer plumbing and fix the comments. As-is it's a protocol field nobody consumes plus a misleading contract.

fuziontech:
## Code review (high effort — 7 finder angles + verification pass)

Self-review via `/code-review`; CI note at the bottom. Five findings are inline on the diff. Four more are architectural and verified against the code but don't sit on a single changed line:

### 1. Flight SQL ingress bypasses interception entirely — silent credential loss *(confirmed, high)*
The interception lives only in the PG-wire conn layer (`handleQuery`/`handleExecute`). A Flight SQL ingress client running `CREATE PERSISTENT SECRET` executes it on the worker (success), it's never persisted to the store, and the next session's mandatory wipe **deletes it** — silent loss with zero diagnostics, and strictly worse than pre-PR (where it at least survived hot-idle reuse). Flight sessions *do* get replay (the loader hangs off the org SessionManager), so the asymmetry is invisible until the secret vanishes. Minimum fix: reject persistent-secret DDL on the Flight path with a clear error; better: hoist interception to a shared execution layer all ingresses traverse.

### 2. Multi-statement simple-query batches: same silent-loss shape *(confirmed)*
`Classify` deliberately refuses multi-statement strings, but `isDuckDBUtilityCommand` still prefix-matches `CREATE PERSISTENT SECRET s (…); SELECT 1`, so the whole batch executes natively on the worker: secret created session-side, success returned, nothing persisted, wiped at next session. The PR's own contract is that persistence failures must surface as errors. Since `Classify` already detects "secret DDL head + trailing statement", the interceptor could reject that combination explicitly instead of falling through.

### 3. Per-org image pinning makes the version-skew window permanent *(confirmed)*
`workerImageForOrg` lets an org run an arbitrarily old worker image indefinitely. An old-image worker silently ignores `secret_statements` (unknown JSON field), returns no `secret_warnings`, and **lacks the create-time wipe** — so for pinned orgs: no replay (silent missing secrets) *and* persistent secrets created via the new CP linger across users of the org on hot-idle reuse — the exact exposure the wipe prevents. There's no version/capability check on the CreateSession RPC. Suggest: worker response advertises secret-replay support (or echoes the count applied); CP treats absence as a loud error/metric, and the release notes should call out the minimum worker image for orgs with pinned images.

### 4. Residual unredacted log site: Describe *(confirmed)*
`conn_extended_query.go:231` logs the raw prepared statement (`"Describe statement." … ps.query`) without `RedactForLog`. Drivers routinely Describe before Execute, so extended-protocol secret DDL writes the full credential pair into debug logs. Same treatment as the Parse/Execute sites needed. (Related inline comment on `redact.go`: consider one redaction chokepoint instead of N call sites — the double-redaction in `execUserSecretDDL` + helpers shows two layers already both believe they own it.)

### Cleanup (non-blocking)
- `usersecrets.skipCommentsAndSpace` re-implements `sqlcore.StripLeadingComments` with different semantics (nested comments, unterminated-comment handling); `quoteSecretIdent` is the ~4th copy of `sqlcore.QuoteIdentifier`; `reservedNames` hardcodes names that exist as constants/literals at their creation sites (`iceberg.LakekeeperSecretName`) — drift here is user-shadowable system secrets.
- `execUserSecretDDL` hand-copies `executeQueryDirect`'s error handling but drops `recoverAbortedTransaction`, so secret DDL behaves differently on an aborted-txn connection than every other statement.
- `Statement.OrReplace` is write-only; `handleUserSecretDDLSimple/Extended` differ only by ReadyForQuery; the harness `_pg_exec`/`pg_try` duplication forced this PR to make the same edit twice.
- Hot path: `RedactForLog` now runs ~4–6× per query across `loggableQuery` + the log helpers; a cheap `EqualFold(head[:6], "CREATE")` pre-check and/or redact-once-per-statement would eliminate per-query `ToUpper` allocations. The user-secret loader is also a serial config-store round trip on every session create — could overlap worker acquisition.

### CI status
- All green except `lint`, fixed in ee02ab1 (two trivial findings in new code); e2e-mw-dev runs on this PR's workflow as usual.

My take on priority: inline DROP-fallback + redaction-mismatch, and #1/#2 above (the silent-loss pair) are worth fixing before merge; #3 needs at least a release-note/metric decision; the rest can follow.

server/conn_user_secrets.go
fuziontech: Fixed in 4c97f31: fallback now gated on `isSecretNotFoundError` (DuckDB's "Failed to remove non-existent secret" message, verified against the pinned engine). Cancel/RPC/ambiguity/aborted-txn errors surface unchanged and never touch the store; regression test `TestExecUserSecretDDLDropOtherErrorsDoNotTouchStore` covers all three classes.

server/conn_user_secrets.go
fuziontech: Fixed in 4c97f31: `Statement.IfNotExists` flows through `PutSecret` to an `ON CONFLICT DO NOTHING` upsert, so an already-stored name is left untouched — matching DuckDB's live-session no-op. Covered by `TestExecUserSecretDDLIfNotExistsFlag`.

server/usersecrets/redact.go
fuziontech: Fixed in 4c97f31: `RedactForLog` is now driven by the same `parseSecretDDLHead` tokenizer as `Classify` — the prefix list is gone, so the redactor structurally cannot fall out of sync with the interceptor. Whitespace/comment-between-keywords and multi-statement cases added to the tests. Bonus: the non-matching fast path is two case-folded keyword compares, no allocation, which also addresses the hot-path efficiency note.

server/conn_extended_query.go
fuziontech: Fixed in 4c97f31: parameterized secret DDL is rejected with 0A000 ("must use literal values ... so they can be replayed") before execution — executing unbound was wrong, and placeholder text could never replay. Test added to the rejection table.

duckdbservice/service.go
fuziontech: Fixed in 4c97f31: the one-session precondition is asserted at the point of reliance — `sharedWarmMode && maxSessions != 1` skips wipe+replay with an ERROR log and surfaces a NOT-restored warning, instead of trusting the k8s_pool flag two layers away.

controlplane/session_mgr.go
fuziontech: Addressed in 4c97f31: the doc comments no longer claim client-facing delivery (warnings are logged worker-side and CP-side with pid/worker context). The plumbing now earns its keep beyond logging: the worker always returns `secret_warnings` (empty, never absent), and the CP treats an absent field with statements sent as an old-image worker — surfacing "persistent secrets were NOT restored" instead of silence. NoticeResponse delivery at session start left as a follow-up.

PR #803, opened 2026-06-23
[codex] Add tenant query log routing and retention

bill-ph:
## Summary

- Route multitenant query logs to each tenant's DuckLake-backed `system.query_log` instead of relying on a process-global metadata store.
- Add query-log retention config, a config-store migration, and a leader-only janitor retention hook.
- Harden query-log provider behavior so logging stays best-effort: nonblocking cache misses, typed-nil handling, single-connection logger DBs, stale profiling cleanup, and self-query skip ordering.
- Update docs/tests for the durable query-log fields and retention defaults.

## Why

The per-query CPU and peak-memory fields only help where durable query logging actually exists. In multitenant mode the control plane may not have a global DuckLake metadata store, so query logging needs to resolve and cache the tenant DuckLake runtime per org. Retention also needs to operate per org from the leader path rather than from each query logger flush loop.

## Validation

- `just test-unit`
- `just test-controlplane`
- `just test-controlplane-k8s`
- `just test-configstore-integration`
- `just lint`


bill-ph:
Closing as obsolete. This was the older tenant DuckLake query-log routing/retention approach; PR #804 is now the active Kafka query-log sink foundation path.

PR #832, opened 2026-06-30
janitor: reclaim hot workers abandoned by a draining/dead CP

fuziontech:
## What

Closes a worker-leak gap that pins idle vCPU/memory. A worker stuck in `hot` with **no session** fell through both cleanup paths:
- the **hot-idle TTL reaper** only reaps `hot_idle`;
- the **stranded-pod reconciler** only deletes `Retired`/`Lost` rows;
- `ListOrphanedWorkers` reaped `hot` workers only once their owning CP row was marked **Expired** — but with the remote backend's **unbounded handover-drain timeout**, a draining CP lingers (up to the pod grace period), so its idle hot workers (no longer health-checked → heartbeat goes stale) accumulate.

**Found live in prod-us:** 132 `hot`, 0-session workers for a single org (`4dc8564d`) reserving **~2,385 vCPU / ~25 TB**, owned across 3 CP replicaset generations — i.e. workers abandoned by CPs that rolled. (This is the residual cause of the long-standing high-vCPU problem; removing the org's size override earlier shrank new workers but never reclaimed the leaked count.)

## Fix

Adds orphan **condition (4)** to `ListOrphanedWorkers`: a worker in a cleanup state whose owner CP is **draining** (or itself heartbeat-stale past the orphan cutoff) **AND** whose own heartbeat is stale past a generous `abandonedWorkerGrace` (5m default) is reclaimed — independent of whether/when the CP row is marked Expired.

Safety guards:
- `abandonedWorkerGrace` is far beyond the worker health-check interval, so a worker **with an active session** — which a draining CP keeps health-checking per the drain contract — keeps a fresh heartbeat and is **never reaped**;
- the existing **reclaimable-Flight-session** exclusion (`NOT EXISTS active/reconnecting session`) is a second line of defense.

Threaded through `ListOrphanedWorkers(before, workerStale)`, the snapshot wrapper, the leader janitor, and the per-CP fallback reaper.

## Tests

New configstore **Postgres** test `TestListOrphanedWorkersReclaimsAbandonedHotWorker` (ran green against real PG): the abandoned hot worker is reclaimed; a **fresh-heartbeat** hot worker and a hot worker owned by a **healthy active CP** are both spared. Existing janitor / per-CP reaper unit tests pass.

**e2e:** not asserted in `tests/e2e-mw-dev` — it needs a draining-CP + stuck-hot-worker fixture the in-Job harness can't construct deterministically; the Postgres + unit tests cover the logic. (Flagging per the harness-coverage rule.)

## Ops note

The immediate prod leak was already mitigated out-of-band (124 idle hot pods for the org deleted, ~2.2k vCPU reclaimed, busy workers preserved). This PR prevents recurrence.

### DIFF
--- a/controlplane/configstore/lifecycle_store.go
+++ b/controlplane/configstore/lifecycle_store.go
@@ -23,8 +23,8 @@ func (cs *ConfigStore) ObserveWorker(workerID int) (*WorkerSnapshot, error) {
 // snapshot binds the observation (state, owner, epoch, updated_at) to
 // the subsequent retire CAS, so callers cannot accidentally retire a
 // row that has changed under them since the listing.
-func (cs *ConfigStore) ListOrphanedWorkerSnapshots(before time.Time) ([]WorkerSnapshot, error) {
-	records, err := cs.ListOrphanedWorkers(before)
+func (cs *ConfigStore) ListOrphanedWorkerSnapshots(before, workerStale time.Time) ([]WorkerSnapshot, error) {
+	records, err := cs.ListOrphanedWorkers(before, workerStale)
 	if err != nil {
 		return nil, err
 	}
--- a/controlplane/configstore/store.go
+++ b/controlplane/configstore/store.go
@@ -1425,6 +1425,18 @@ func (cs *ConfigStore) countOrgAdmittingWorkers(tx *gorm.DB, orgID, image, profi
 //  3. owner_cp_instance_id is set but no matching cp_instances row
 //     exists at all (hard-deleted somehow), and again the heartbeat
 //     is stale. Same shape as (2), different cause.
+//  4. The owning CP row exists and is DRAINING (or itself heartbeat-stale
+//     past `before`) AND the worker's own heartbeat has gone stale past
+//     `workerStale`. This closes the leak where a CP drains (or dies
+//     ungracefully) without being marked expired — with the remote
+//     backend's unbounded handover-drain timeout a draining CP can linger
+//     for the full pod grace period, and its idle `hot` workers (no longer
+//     health-checked, so their heartbeat goes stale) were invisible to
+//     (1)/(2)/(3). `workerStale` is a deliberately generous grace (well
+//     beyond the health-check interval) so a worker with an active session
+//     — which a draining CP keeps health-checking per the drain contract —
+//     keeps a fresh heartbeat and is never reaped here; the Flight-session
+//     guard below is a second line of defense.
 //
 // Retired/lost rows are deliberately excluded — their pods are already
 // gone (or are reconciled by K8sWorkerPool.cleanupOrphanedWorkerPods).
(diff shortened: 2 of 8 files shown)


fuziontech:
## ⚠️ Self-review (adversarial) found this approach is broken — converted to draft, do not merge

An independent deep review caught two interlocking flaws I missed:

**1. CRITICAL — the safety premise is false.** The PR assumes "a draining CP keeps health-checking active-session workers, so their heartbeat stays fresh and they're never reaped." That's wrong: `last_heartbeat_at` is written **only on state transitions** (`k8s_pool_helpers.go` `workerRecordFor`) + the activating-heartbeat loop (which stops the moment the worker commits `Hot`). The health-check **success path never writes the worker record**. So `last_heartbeat_at` is a *last-state-transition* timestamp, not a liveness signal — a worker running a long (>5m) PG query has a stale heartbeat. Condition (4) therefore **lists a live-query worker** (draining owner + stale worker hb), and PG-wire sessions aren't in `flight_session_records`, so that guard doesn't protect it. My test case modeling "fresh heartbeat = active session" is unreachable.

**2. HIGH — it's also inert.** `RetireOrphanWorker` re-validates at CAS time and only retires when the owner CP is **expired** (`NOT EXISTS cp WHERE state <> expired`). Condition (4) targets **draining** owners (state ≠ expired) → every retire CAS misses → no pod delete → **the leak is not fixed**. Proven by the existing `TestRetireOrphanWorkerRejectsRevivedOwnerControlPlane`. My new test only asserts on the LIST, never the retire, so it passed green over a no-op.

Net: the change is simultaneously ineffective **and** one fence-relaxation away from killing customer queries.

**3. MEDIUM** — the `cp.last_heartbeat_at <= 30s` arm is dead weight (leader expires CPs at 20s, so condition (1) already owns that case) and a latent false-positive if the fence were relaxed.

### Path forward
The prod symptom was already mitigated out-of-band (idle pods deleted). The durable fix needs a **real worker-liveness/idle signal** before any reaping of draining-CP-owned workers — e.g. periodically re-stamp `last_heartbeat_at` for Hot/HotIdle workers, or gate the reap on worker-reported `activeSessions==0` — **and** a matching relaxation of the `RetireOrphanWorker` fence, with a *retire-path* test (not just a list test) + an e2e harness assertion. Alternatively, fix the leak at the source (a draining CP should release its idle hot workers to `hot_idle`, or get force-expired once its workers are idle) so the existing correctly-fenced path reclaims them. Reworking on a fresh branch.

fuziontech:
Closing in favor of #835. This PR fixed the stuck-Hot leak on the *reaper* side, but review found it broken: the safety premise (a draining CP keeps `last_heartbeat_at` fresh on active-session workers) is false — `last_heartbeat_at` is only written on state transitions, not on health-check success — so the new `ListOrphanedWorkers` condition would list live-query workers; and `RetireOrphanWorker` only retires when the owner CP is `expired`, never `draining`, so the condition was also inert against its own target. #835 fixes it at the source instead: a draining CP parks its idle Hot workers to `hot_idle` at drain start, moving them into the unfenced hot-idle TTL reaper path.

PR #834, opened 2026-06-30
admin: aggregate live session/query state across CP replicas (fix flashing)

fuziontech:
## What

The Live view (running queries + sessions) and the Overview session counts **flickered between real values and 0** on every 3s poll. Root cause: the live session/query endpoints read each CP's **in-memory** view (a session lives only on the CP that owns it). Behind the load-balancer + 6 CP replicas, each poll landed on a random pod and showed only that pod's slice — so e.g. 4 data-import queries on one CP flashed `4 ↔ 0` as the ALB bounced between pods.

## Fix — HTTP fan-out (pull)

The serving CP:
1. discovers its peer CP pods (K8s pod list, matched by the shared `duckgres-control-plane-` name prefix — worker/cache pods are excluded),
2. GETs each peer's `?scope=local` view (the **recursion guard** — a peer returns only its own slice, never re-fans) authenticated with the internal secret,
3. concatenates (a session/worker is owned by exactly one CP → the union is **disjoint**, no dedup).

Applied to **`/queries`, `/sessions`, `/workers`, `/status`**. Peers are fetched **concurrently** with a short per-peer timeout; a slow/down peer is omitted (graceful degradation), and `/queries` returns `cp_responders`/`cp_total` so the UI can show coverage. `/workers/fleet` is already cluster-wide (config store) and isn't fanned out.

### Why fan-out, not a message bus
The live data is **authoritative in each CP's memory** and only read while an operator has the dashboard open. A handful of on-demand GETs across ~6 pods fits that far better than a ZeroMQ/pub-sub mesh (which would add libzmq/CGO, a persistent connection mesh, and stale-replica/eviction logic for a 3s poll). `PeerFetcher` is `nil` in single-CP / test setups (local-only behavior preserved).

## Tests
- `TestQueriesAggregateAcrossCPs` / `TestSessionsAggregateAcrossCPs`: merge local+peer, and assert `?scope=local` does **not** fan out.
- e2e `admin_console_api` now asserts `/queries` returns `cp_total >= 1` (the fan-out is wired in-cluster).
- **Frontend needs no change** — it reads the same `{queries}` envelope / session array; the added meta fields are ignored.

Closes the last item from the prod-review thread (the "flashing" #4).

### DIFF
--- a/controlplane/admin/README.md
+++ b/controlplane/admin/README.md
@@ -62,6 +62,21 @@ Added for the console:
 | `POST /api/v1/orgs/:id/impersonate/query` | admin | run SQL as an org user on their worker |
 | `GET /api/v1/audit` | admin | admin action log |
 
+### Cross-CP live-state aggregation (`live_aggregate.go` + `controlplane/live_aggregator.go`)
+
+Live session/query state is **in-memory per CP** — each replica only knows the
+sessions it owns. Behind the load-balancer that made the dashboard's numbers
+flicker as polls landed on different pods. The session/query endpoints
+(`/queries`, `/sessions`, `/workers`, `/status`) **fan out**: the serving CP
+discovers its peer CP pods (K8s pod list, name-prefix match), GETs each peer's
+`?scope=local` view (the recursion guard — a peer returns only its own slice)
+with the internal secret, and concatenates (a session is owned by exactly one
+CP, so the union is disjoint — no dedup). Peers are fetched concurrently with a
+short per-peer timeout; a slow/down peer is omitted, and `/queries` reports
+`cp_responders`/`cp_total` for coverage. `PeerFetcher` is nil in single-CP /
+test setups (local-only). `/workers/fleet` is already cluster-wide (config
+store) and is not fanned out.
+
 ### Impersonation (`impersonate.go` + `controlplane/admin_providers.go`)
 
 `POST /api/v1/orgs/:id/impersonate/query` `{username, sql, allow_write}` opens a
--- a/controlplane/admin/api.go
+++ b/controlplane/admin/api.go
@@ -1013,10 +1013,11 @@ func (h *apiHandler) listWorkers(c *gin.Context) {
 	if h.info != nil {
 		workers = h.info.AllWorkerStatuses()
 	}
-	// A worker is owned by exactly one CP, so the union across CPs is disjoint.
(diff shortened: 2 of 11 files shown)


fuziontech:
## Adversarial self-review + fixes (pushed `752c921`)

An independent deep review verified the core design as sound — the `scope=local` recursion guard is real and total, the internal-secret peer hop only reaches the same viewer-readable read paths (no privilege escalation), merge is panic-safe on garbage bodies, concurrency is locked, and single-CP/nil-fetcher degrades to local-only. Findings addressed:

- **HIGH — self-double-count.** Self was excluded only by pod name (`== cpID`); if `DUCKGRES_K8S_CONTROL_PLANE_ID` overrides `cpID` to a non-pod-name value, the serving CP would fan out to itself and ~2x its own numbers. **Fixed:** also exclude self by IP (`POD_IP` env + own interface addrs), and the merge is now idempotent via `dedupeBy(worker_id)` on `/queries`/`/sessions`/`/workers` (which also closes the takeover/handover double-count window).
- **MEDIUM — silent discovery failure.** `discoverPeerIPs` now logs (debug) when it finds zero peers, so a chart-label/RBAC mismatch is diagnosable rather than a silent revert to per-CP flicker.
- **MEDIUM — gofmt** break in the test file: fixed (was failing the fmt gate).
- **MEDIUM — thin tests:** added `TestDiscoverPeerIPs` (fake clientset: self-by-name + self-by-IP excluded, cross-replicaset peer included, non-CP/pending/no-IP skipped), `TestCPDeploymentPrefix`, `TestStatusAggregateAcrossCPs`, `TestAggregateSkipsBadPeer`, `TestAggregateDedupesByWorker`.
- **MEDIUM — tautological e2e assertion:** the CI CP is single-replica so it can't exercise a real peer hop; the assertion now validates the coverage envelope (`cp_total == cp_responders`) and the multi-CP path is covered by the unit tests above (documented per the harness-coverage rule).
- **LOW — doc:** corrected the `mergeOrgStats` comment (`OrgStatus.Workers` is 0 from the adapter today).

Acceptable as-is (noted, not changed): the internal secret traverses plaintext pod-to-pod HTTP on a 3s cadence (consistent with the existing ALB→pod posture, Cilium-scoped), and there's no caching/singleflight (bounded: authed operators only; ~`4×(N−1)` in-memory reads per poll).

PR #838, opened 2026-06-30
admin/live: per-query detail view (redacted SQL + conn metadata)

fuziontech:
## What

Clicking a running-query row on the admin **Live** page now opens a detail dialog with the query's redacted SQL text, connection metadata (worker pod, client addr, application, elapsed), and live progress. It's fetched on demand via `GET /api/v1/queries/:pid` so the 3s-polling list payload stays light.

This is phase 1 of the live-query detail idea — the redacted-SQL + elapsed + progress view that gets most of the operational value. Phase 2 (operator-level DuckDB plan/progress surfaced over Flight) is deliberately out of scope.

## How it's safe

The SQL comes from the existing per-connection `currentQuery`, which is **already redacted** through `usersecrets.RedactForLog` (the exact value `pg_stat_activity` exposes). Credentials in `CREATE SECRET` statements never reach the admin UI. There is no path here that reads raw SQL.

## Scope caveat

The SQL text only exists in-process on the CP replica that owns the client connection, so `/queries/:pid` returns **404** for a query owned by a *different* replica (rather than empty text). This matches the existing `/queries` list, which is also local-stack-scoped. Full cross-replica detail would be a follow-up fanout.

## Changes

- **server**: `ConnDetail` + `Server.ConnDetailByPID` — redacted conn snapshot.
- **admin**: `QueryDetail` type, `LiveInfo.QueryDetailForPID`, `GET /api/v1/queries/:pid` (400 bad pid / 404 not-owned).
- **controlplane**: provider joins the conn snapshot with the session manager's protocol + progress; `srv` wired into the provider.
- **ui**: new `QueryDetailDialog` + `useQueryDetail` hook (polls while open, stops on 404); Live query rows are clickable (Cancel button `stopPropagation`s).
- **docs**: admin README endpoint table + `docs/design/admin-ui.md`.

## Tests

- `server/conn_detail_test.go` — asserts the **redaction guarantee** (credential material never appears in the snapshot), state derivation, and missing/nil-pid handling.
- `controlplane/admin/live_test.go` — route returns 200 / 404 / 400 correctly.
- `tests/e2e-mw-dev/harness.sh` — new `admin_query_detail` assertion: runs a real query on a warm cnpg worker, finds its pid via `/queries`, expands it and verifies the SQL round-trips + identity matches, and that an unknown pid 404s.

### DIFF
--- a/controlplane/admin/README.md
+++ b/controlplane/admin/README.md
@@ -60,7 +60,7 @@ Added for the console:
 |-------|------|---------|
 | `GET /api/v1/me` | any | caller identity + role (SPA tailors its UI) |
 | `GET /api/v1/queries` | viewer | running queries w/ progress, `?org=&user=` slicing |
-| `GET /api/v1/queries/:pid` | viewer | one query's detail: redacted SQL text + conn metadata + progress. Scatter-gathers like `/queries` — checks locally, else fans out to peer CPs (`?scope=local` guard); 404 only if no replica owns the pid |
+| `GET /api/v1/queries/by-worker/:wid` | viewer | one query's detail: redacted SQL text + conn metadata + progress, addressed by cluster-unique worker id (pid is per-org, not unique). Scatter-gathers like `/queries` — checks locally, else fans out to peer CPs (`?scope=local` guard); 404 only if no replica owns the worker |
 | `GET /api/v1/sessions`, `/workers` | viewer | live sessions / session-holding workers |
 | `GET /api/v1/workers/fleet` | viewer | cluster worker counts by lifecycle state |
 | `GET /api/v1/cluster/instances` | viewer | live CP replicas (self-flagged) |
--- a/controlplane/admin/README.md
+++ b/controlplane/admin/README.md
@@ -53,6 +53,7 @@ Added for the console:
 |-------|------|---------|
 | `GET /api/v1/me` | any | caller identity + role (SPA tailors its UI) |
 | `GET /api/v1/queries` | viewer | running queries w/ progress, `?org=&user=` slicing |
+| `GET /api/v1/queries/:pid` | viewer | one query's detail: redacted SQL text + conn metadata + progress (scoped to the owning CP replica; 404 otherwise) |
 | `GET /api/v1/sessions`, `/workers` | viewer | live sessions / session-holding workers |
 | `GET /api/v1/workers/fleet` | viewer | cluster worker counts by lifecycle state |
 | `GET /api/v1/cluster/instances` | viewer | live CP replicas (self-flagged) |
--- a/controlplane/admin/live.go
+++ b/controlplane/admin/live.go
@@ -77,10 +77,13 @@ type CPInstance struct {
 type LiveInfo interface {
 	// RunningQueries returns every active session with its cached progress.
 	RunningQueries() []QueryStatus
-	// QueryDetailForPID returns the expanded detail (redacted SQL + connection
-	// metadata + progress) for one in-flight query, or ok=false if no live
-	// session with that pid is owned by this control-plane replica.
-	QueryDetailForPID(pid int32) (QueryDetail, bool)
+	// QueryDetailForWorkerID returns the expanded detail (redacted SQL +
+	// connection metadata + progress) for the in-flight query on the given
+	// cluster-unique worker id, or ok=false if no live session/conn for that
+	// worker is owned by this control-plane replica. Addressed by worker id, not
+	// pid: pids are per-org and not unique, so a pid key can return the wrong
+	// org's query.
+	QueryDetailForWorkerID(workerID int) (QueryDetail, bool)
 	// WorkerFleet returns worker counts by lifecycle state across the cluster.
 	WorkerFleet() ([]FleetStat, error)
 	// ControlPlaneInstances returns the live CP replicas.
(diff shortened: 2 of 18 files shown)


fuziontech:
## Adversarial self-review

Ran an independent reviewer over the diff with a focus on the load-bearing redaction invariant, concurrency, authz, and test honesty. **No blockers** — summary below, with the cheap high-value items already addressed in a follow-up commit.

### ✅ Redaction (priority 1) — verified solid
- `currentQuery` is written at exactly two sites (`conn.go:1170`, `conn_extended_query.go:506`), **both** storing `usersecrets.RedactForLog(...)`. `connDetail()` reads only that already-redacted value and never touches `p.stmt.query` / raw text.
- The detail payload has **no error-message field**, so the "engine echoes the offending SQL in error text" leak vector isn't reachable here.
- Hot-idle worker reuse can't cross-leak: `currentQuery` lives on the per-connection `clientConn` keyed by PID, not on the shared worker.

### ✅ Confirmed fine
- **Route:** `GET /queries` + `GET /queries/:pid` don't conflict in gin; PID parse is robust (400 on garbage/overflow, 404 on unknown/negative). `TestQueryDetailRoute` covers 200/404/400.
- **Authz:** endpoint sits under `AuthMiddleware` + `RoleGate`; GET is viewer-tier by design (documented), no mutation slipped in, Cancel still routes through the admin-gated `POST /sessions/:pid/cancel`.
- **Nil guards:** nil `*Server`, nil `srv`, nil `c.conn` all handled and unit-tested. `connDetail` test passes under `-race`.
- **UI:** 404 detection works; `refetchInterval` stops on error and `enabled` disables polling when closed (no leak); `stopPropagation` on row Cancel correct; per-pid queryKey avoids stale cross-pid data.

### Addressed in follow-up commit
- **O(1) protocol lookup** — added `SessionManager.ProtocolForPID`, replacing an `AllSessions()` full-map copy in `QueryDetailForPID`.
- **Live-path redaction probe** — `harness.sh` now runs a real `CREATE SECRET` on a worker, catches it in flight via `/queries/:pid`, and asserts the credential material is **absent** (statement still shows, redacted). Closes the loop the unit test couldn't (it only proved pass-through of a pre-redacted value).

### Known limitation (documented, not fixed here)
- **Multi-replica scoping:** the SQL text only exists on the CP replica that owns the connection. With >1 CP replica behind the internal ALB and no session affinity, the list (replica A) and the detail fetch (replica B) can disagree → a 404 for a genuinely-running query. This matches the existing `/queries` list's per-replica behavior. Proper fix is session stickiness on the admin ALB (`ingress-admin.yaml`) or a cross-replica fanout — tracking as a follow-up; the dialog copy already explains the 404 case.

🤖 review by an independent agent + author verification

fuziontech:
### Update: multi-replica detail now scatter-gathers

Rebased onto main's cross-CP live-state fan-out (#834) and wired `/queries/:pid` into it, resolving the multi-replica limitation called out in the review above.

`/queries/:pid` now behaves exactly like `/queries`: it checks the local replica first, and if this CP doesn't own the pid it fans out to peer CPs via the `PeerFetcher` (each peer answers `?scope=local` — the recursion guard — and `FetchPeers` drops every peer's 404, so only the owning replica's 200 is returned). A 404 now means *no* replica owns the pid (the query ended), not "the wrong replica answered." No ALB stickiness required.

- `TestQueryDetailFansOutAcrossCPs` asserts the peer merge **and** the `scope=local` no-fan-out guard.
- Dialog 404 copy + README + design doc updated to match.

fuziontech:
### Added: running-query duration in the Live view

Per request, the Live table now has a **Duration** column showing how long each in-flight statement has been running (not just progress).

- `QueryStatus.ElapsedMS` — computed on the owning CP from the connection's query-start (`ConnDetailByPID`), so it's per-item and survives the cross-CP `/queries` merge unchanged; 0 when idle.
- Shared `fmtDuration()` in `lib/format.ts` now backs both the new column and the detail dialog (replaced the dialog's local `fmtElapsed`).
- `harness.sh admin_query_detail` asserts the in-flight query's list item reports a positive `elapsed_ms`.

fuziontech:
## Adversarial review — BLOCKER found and fixed

An independent reviewer found a real correctness **BLOCKER**, now fixed in the latest commit.

**The bug:** backend PIDs are allocated **per-org** — every stack's `SessionManager` starts at 1000 (`session_mgr.go`) — so two orgs (within one CP *or* across replicas) can share a pid, and the shared `cp.srv.conns` map is keyed by pid. Addressing `/queries/:pid` by pid could return the **wrong org's query**: a non-owner CP with a same-valued pid would short-circuit locally, and `QueryDetailForPID` re-scanned all stacks for the first pid match (random map order), stitching one org's SQL onto another's identity.

**The fix:** address detail by the **cluster-unique worker id** (config-store issued — the same key the list already dedupes on):
- `GET /api/v1/queries/by-worker/:wid`; `ConnDetailByWorkerID` / `SessionForWorker` / `QueryStartsByWorkerID`; UI dialog/hook/api/row all key on `worker_id`.
- New `TestQueryDetailWorkerIDDistinguishesCollidingPIDs`: two queries sharing pid 1001 on different workers/orgs each resolve correctly.
- Also applied the reviewer's other items: atomic `callCount()` on the fake fetcher; the harness redaction probe now pins to the actual `CREATE SECRET` statement (via the detail SQL) instead of grabbing any org pid.

**Verified clean by the reviewer:** redaction (only the already-redacted `currentQuery` is read; credential-absent unit + e2e asserts hold), ElapsedMS merge-safety, the `scope=local` recursion guard, concurrency/nil-guards, and authz (GET viewer-tier, no mutation added).

### Flagged, pre-existing, NOT fixed here (own follow-up)
The conns map is keyed by the per-org pid, so colliding conns **shadow each other** — this also affects `pg_stat_activity` and cancel-by-pid (`KillSession`). With worker-id addressing the detail now degrades to **404** for a shadowed conn rather than showing wrong data, but the root fix (globally-unique CP pids, or re-keying the conns map) is a separate change. Cancel-by-pid remains pid-based.

PR #846, opened 2026-06-30
admin/ui: frontend test suite + Overview worker-count derivation refactor

fuziontech:
## Why

The admin dashboard's data-derivation logic has shipped wrong more than once — the Workers card showing the **same number for hot and idle**, and a leak warning firing **while every worker was busy** (the prod screenshot below). Root cause beyond the specific bugs: **there were zero frontend tests**, and the math lived inline in JSX where nothing could pin it.

> ⚠ 150 hot workers are not holding a session — possible idle/leaked workers reserving vCPU & memory.

(That exact case — 150 `hot`, 150 sessions, 0 `hot_idle` — is now a regression test.)

## What

- **Frontend test infrastructure** (first in the repo): Vitest + Testing Library + jsdom. `npm run test`, `just ui-test`, and a new CI `ui-tests` job (typecheck + unit tests on every PR). Previously the UI was only ever type-checked inside the Docker build.
- **Extracted the bug-prone math out of `Overview.tsx`** into pure, testable modules (`src/lib/fleet.ts`): `summarizeFleet` (busy=`hot` vs idle=`hot_idle`, total, by-state), `isIdleLeak` (warning threshold), `orgLoadPercent`, `topOrgsByLoad`. Derivations no longer live in the JSX.
- **`src/lib/fleet.test.ts`** (13 tests) pins all of it, incl. the 150-hot/0-idle regression (busy≠idle by construction; no false warning).
- **`src/pages/Overview.test.tsx`** renders the page with mocked hooks and asserts the Workers card text + the leak-warning presence/absence at the render level.

## Build safety

Production build unchanged: `npm ci` (Docker) installs the test devDeps so `tsc -b` has vitest types; test files aren't bundled. Verified `npm run build`, `typecheck`, `lint` (0 new warnings), and `just ui-test` all green.

## Related follow-up (not in this PR)

The per-org load bars still render empty because the backend leaves `OrgStatus.Workers` at 0 (flagged in #839). `orgLoadPercent` is covered here, but the real fix is a separate backend change to populate per-org worker counts. Happy to take that next.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,24 @@ jobs:
       - name: Run cache-proxy tests
         run: just test-cache-proxy
 
+  ui-tests:
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Set up just
+        uses: extractions/setup-just@dd310ad5a97d8e7b41793f8ef055398d51ad4de6 # v2.0.0
+
+      - name: Set up Node
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: "20"
+          cache: npm
+          cache-dependency-path: controlplane/admin/ui/package-lock.json
+
+      - name: Run admin UI typecheck + unit tests
+        run: just ui-test
+
   integration-tests:
     needs: unit-tests
     runs-on: ubuntu-24.04-arm
--- a/controlplane/admin/README.md
+++ b/controlplane/admin/README.md
@@ -128,8 +128,18 @@ embedded, under Vite, or under the Go devserver.
 
 ## Tests
 
-`authz_test.go` (SSO role mapping, RoleGate, SQL classifier), `dashboard_test.go`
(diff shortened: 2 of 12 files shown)


fuziontech:
## Adversarial review — clean, no blockers

A reviewer ran the full Docker-equivalent gauntlet (`npm ci && typecheck && test && build && lint`, node 20/22) and verified:
- **Production build safety**: no test files leak into `dist/`; `tsc -b` typechecks tests fine because both Dockerfiles `npm ci` (devDeps incl. vitest) before `npm run build`; the `vite.config.ts` `test` block leaves production output unchanged.
- **Behavior preservation**: `fleet.ts` is line-for-line equivalent to the prior inline `Overview.tsx` logic — no off-by-one/NaN/mutation/divide-by-zero.
- **Test quality**: the 150-hot/0-idle/150-session regression is genuinely pinned in both the unit and component tests; the component mocks control hook *inputs* only (not self-referential); CI `ui-tests` job + setup-node SHA verified.

Three NITs, all addressed in the latest commit:
1. `Overview.test.tsx` selected cards by the brittle `.p-4` class → `StatCard` now has a stable `data-testid="stat-<label>"` and the test uses it.
2. "never equal by construction" overstated (5 hot + 5 hot_idle *can* coincide) → reworded to "distinct states, not derived from each other".
3. `just ui-test` (the CI job) now also runs `npm run build`, gating the embedded SPA build on every PR.

PR #847, opened 2026-06-30
admin: populate per-org worker count (fix always-empty load bars)

fuziontech:
## The bug

`OrgStatus.Workers` is declared in the admin API but **never populated** — `AllOrgStats` only set `Name`/`ActiveSessions`/`MaxWorkers`. So:
- the Overview **per-org load bars** always rendered empty (`0/∞ wk`, 0% fill), and
- `ClusterStatus.total_workers` was always 0 (the same dead field that made the Workers card derive a bogus idle count before #839).

## Fix

Populate it from each org's `OrgReservedPool`:
- New `OrgReservedPool.WorkerCount()` returns the org's **cap-counting** assigned workers (`Hot`; hot-idle excluded — same rule as `atOrgWorkerCap`).
- `AllOrgStats` sets `OrgStatus.Workers` from it via a type assertion (remote/k8s mode → `*OrgReservedPool`).

It's a **per-CP local** count — a worker is owned by exactly one CP — and the `/status` fan-out (`mergeOrgStats`) already **sums** `Workers` across replicas, so the cluster-wide per-org total comes out right, consistent with how `active_sessions` is merged. `total_workers` becomes the real cluster total as a bonus.

## Tests

- **`TestOrgReservedPoolWorkerCount`**: counts the org's Hot workers, excludes hot-idle and other-org workers sharing the pool, and returns 0 for an org with none.
- **`harness.sh admin_per_org_workers`**: with a query in flight, `/status` reports `workers>=1` for that org and `total_workers>=1` (proves the field is live end-to-end, not 0).

Completes the per-org follow-up noted in #839 / #846.

### DIFF
--- a/controlplane/admin/live_aggregate.go
+++ b/controlplane/admin/live_aggregate.go
@@ -59,11 +59,11 @@ func dedupeBy[T any, K comparable](items []T, key func(T) K) []T {
 }
 
 // mergeOrgStats folds each peer's ClusterStatus into the local per-org stats:
-// active_sessions is summed across CPs (each reports its own slice). Workers is
-// also summed for forward-compatibility, but the adapter currently leaves it 0
-// (per-org worker counts aren't surfaced), so it's effectively a no-op today.
-// max_workers is config-store-backed (identical on every CP) so the max is
-// taken. Org identity is the name; order follows first-seen.
+// active_sessions and workers are both summed across CPs (each reports only the
+// sessions / workers it owns — a worker is owned by exactly one CP, so the sum
+// is the cluster-wide per-org count). max_workers is config-store-backed
+// (identical on every CP) so the max is taken. Org identity is the name; order
+// follows first-seen.
 func mergeOrgStats(local []OrgStatus, bodies [][]byte) []OrgStatus {
 	byName := map[string]*OrgStatus{}
 	var order []string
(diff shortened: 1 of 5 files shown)


fuziontech:
## Adversarial review — clean, no blockers

The reviewer focused on the thing most likely to be subtly wrong — **cross-CP double-counting** — and verified it's safe: `WorkerCount()` counts only this CP's local `Hot` (non-hot-idle) workers for the org; the only cross-CP-claimable state is hot-idle (via `ClaimHotIdleWorker`/`TakeOverWorker`), which is excluded on **both** sides, and a Hot worker with a live session can't be snatched. So a worker is counted by exactly one CP and the `/status` sum is correct. Locking matches the adjacent `atOrgWorkerCap()`; the numerator/denominator are self-consistent (both exclude hot-idle); the `*OrgReservedPool` type assertion never falls through in the k8s build; `TestOrgReservedPoolWorkerCount` is solid.

Two NITs, addressed in the latest commit:
1. The `mergeOrgStats` doc comment still said Workers summing was "effectively a no-op today" — this PR makes it live, so the comment now lies. Fixed.
2. Noted the `admin_per_org_workers` harness precondition (relies on a warm org so the probe lands on a Hot worker within the 120s budget).

PR #851, opened 2026-06-30
controlplane: make backend pids process-global (fix conns-map collision)

fuziontech:
## The root cause

Surfaced by the #838 live-query-detail review. Backend pids were allocated **per-org** — every `SessionManager`'s `nextPID` started at 1000 (`controlplane/session_mgr.go`) — but every client connection registers into the **one** `server.conns` map keyed by pid. So two orgs' connections at the same pid value collided in that map:

- the later registrant **shadowed** the earlier, which then vanished from `pg_stat_activity` and every pid-keyed lookup;
- either connection closing (`delete(s.conns, pid)`) removed the **other** org's entry;
- cancel-by-pid (`KillSession` scans stacks for the first matching pid) could tear down the **wrong** org's session.

#838 worked around the read side by addressing detail via the cluster-unique worker id (`/queries/by-worker/:wid`), but the underlying map collision remained for `pg_stat_activity` and cancel.

## The fix

Replace the per-manager counter with a single **process-global** atomic (`globalNextPID`, starts at 1000). Pids are now unique within a CP, so the conns map can't collide — `pg_stat_activity`, cancel-by-pid, and any pid-keyed lookup are correct again.

Cross-CP pids can still coincide, but each CP has its own conns map, and cross-CP addressing already uses the cluster-unique worker id. PG clients don't assume pid ranges, so global pids are transparent to them.

## Tests

- **`TestReservePIDGloballyUniqueAcrossManagers`**: two managers (two org stacks) reserving pids never hand out the same value.
- This is the **deterministic gate** — a real collision can't be forced in the warm e2e cluster (it needs a fresh CP with two orgs opening their first connection at the same pid count), which is noted in `harness.sh`.

Closes the follow-up flagged in #838.

### DIFF
--- a/controlplane/session_mgr.go
+++ b/controlplane/session_mgr.go
@@ -64,6 +64,22 @@ var globalNextPID = func() *atomic.Int32 {
 	return v
 }()
 
+// reservePID returns the next backend pid from counter, skipping 0. Backend pids
+// are int32; after ~2.1B connections in one process the counter wraps and passes
+// through 0, which the session-create path treats as "unset" (re-allocating and
+// shipping a stale pid to the client → broken cancel for that one connection).
+// Skipping 0 closes that wrap-time edge with a single extra Add. Negative values
+// after wrap are still unique within the CP's conns map and are left as-is
+// (reaching them needs a further ~2.1B connections, and a CP restarts on every
+// deploy long before either bound).
+func reservePID(counter *atomic.Int32) int32 {
+	p := counter.Add(1)
+	if p == 0 {
+		p = counter.Add(1)
+	}
+	return p
+}
+
 // SessionManager tracks all active sessions and their worker assignments.
 type SessionManager struct {
 	mu         sync.RWMutex
@@ -168,7 +184,7 @@ func (sm *SessionManager) SetRequestedVCPUsResolver(fn func(profile *WorkerProfi
 
 // ReservePID generates a new unique PID for a session.
 func (sm *SessionManager) ReservePID() int32 {
-	return globalNextPID.Add(1)
+	return reservePID(globalNextPID)
 }
 
 func (sm *SessionManager) acquireSlot(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


fuziontech:
## Deep adversarial review — no blockers; one wrap-edge hardened

Traced every backend-pid allocator and stress-checked the overflow/cross-CP/test-assumption angles. **No BLOCKERs, no SHOULD-FIX.** The invariant the PR claims (pids unique within a CP; cross-CP disambiguated by worker id) holds.

**Verified clean:**
- **Allocator completeness** — `globalNextPID` is the *only* source feeding the CP `cp.srv.conns` map. Standalone mode uses a separate `server.nextPID` (reached only via the standalone accept loop, never `cp.srv`); per-child worker `Server`s have their own conns map; impersonation creates sessions but never registers into `conns`. No other allocator can collide into the CP map.
- **Cross-CP** — every cross-CP merge dedups on `WorkerID` (`live.go:119`, `api.go:1058/1074`), never pid, so cross-CP pid coincidence is irrelevant. Detail is `/queries/by-worker/:wid`.
- **PG wire / cancel** — pid is an opaque int32 in BackendKeyData; cancel looks up `s.conns[pid]` unaffected by value; the temp + real conn share the pid.
- **Tests** — no test assumes "first ReservePID == 1001" (the 1001/1002 literals are explicit inputs). The uniqueness test is robust to the shared monotonic global (`-count=5` stable).
- **Concurrency** — the package-init + `.Add(1)` is race-free.

**Hardened (the one NIT):** int32 wrap (~2.1B connections/process — pre-existing, but now reached sooner under multi-tenant load) had one concrete edge: `ReservePID` could return `0` at the wrap point, which the session-create path treats as "unset" → re-allocates and ships a stale pid → broken cancel for that connection. Latest commit routes allocation through a pure `reservePID()` helper that **skips 0**, with `TestReservePIDSkipsZeroAtWrap`. Negative post-wrap values stay unique within the CP conns map and are left as-is (reaching them needs a further ~2.1B connections).

PR #854, opened 2026-06-30
controlplane: default a 60s connection idle timeout (free idle workers)

fuziontech:
## Why

In control-plane mode **one connection pins one worker** (a scarce k8s pod / local process). A connection that connects and then issues no queries — or sits `idle in transaction` — holds its worker "hot" **forever**. That's the worker accumulation we've been fighting: the Live view shows sessions backend-started 3+ hours ago, `idle` / `(no active statement)`, pinning workers that do nothing.

## The gap

The idle-timeout **mechanism already exists** — the message loop sets a read deadline from `cfg.IdleTimeout`, closes the connection when it fires, and the handler's `defer DestroySession` releases the worker back to **hot-idle**. But:

- `server.New()` applies the default (24h);
- the control plane builds its server via `InitMinimalServer`, which **skips** `New()`'s defaulting;
- so `IdleTimeout` was `0` = **disabled**, and CP connections were never idle-reaped.

## The fix

- `server.NormalizeIdleTimeout(configured, zeroDefault)` — shared resolution (`0` → default, negative → disabled, positive → as-is). `server.New` uses it with the 24h standalone default.
- `server.DefaultControlPlaneIdleTimeout = 60s`; the control plane applies it before `InitMinimalServer`. `--idle-timeout` overrides; a negative value disables.

A connection idle for 60s is now closed → `DestroySession` → **worker returns to hot-idle** (not retired), available for reuse. Standalone is unchanged (24h).

## Tests

- **`TestNormalizeIdleTimeout`** — the resolution table.
- **`TestMessageLoopIdleTimeoutClosesConnection`** — an idle connection (via `net.Pipe`) hits the read deadline and the loop returns a clean close. These two are the deterministic gate.
- **`harness.sh conn_idle_timeout_reaps_session`** — holds a real connection `idle in transaction` past the timeout and asserts its session **vanishes from `/queries` while the client is still connected** (so the reap is the CP's, not the client exiting), end-to-end on the mw-dev cluster.

Note: the semantic is "no client traffic for the timeout" (read-deadline based), which for a worker-pinning idle/idle-in-transaction connection is exactly query-inactivity.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,6 +68,7 @@ Key CLI flags for control-plane mode:
 - `--process-min-workers N` / `--process-max-workers N`
 - `--process-retire-on-session-end`
 - `--worker-queue-timeout DURATION` / `--worker-idle-timeout DURATION`
+- `--idle-timeout DURATION` — connection idle timeout: a client connection with no traffic for this long is closed and its worker released to hot-idle (in control-plane mode an idle connection otherwise pins a worker forever). **Control-plane default is `60s`** (`server.DefaultControlPlaneIdleTimeout`; standalone defaults to `24h`); a negative value disables it. `server.New` applies the standalone default, so the control plane sets it explicitly before `InitMinimalServer` (which skips that defaulting).
 - `--memory-budget SIZE` (default 75% RAM) / `--memory-rebalance`
 - `--socket-dir /path` (process backend)
 - `--handover-drain-timeout DURATION` (default `24h` process; **remote default is `0` = unbounded** — the CP waits for active sessions for as long as it takes and the pod's k8s `terminationGracePeriodSeconds` is the only hard wall. cloudflare/tableflip FD passing applies to process/standalone single-host upgrades, not k8s pod replacement.)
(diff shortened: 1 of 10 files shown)


fuziontech:
## Deep review — one BLOCKER found and fixed; everything else clean

An independent adversarial review + my own pass. **No human reviews were left on the PR** (only the test-impact bot).

### 🛑 BLOCKER (fixed in the latest commit)
Lowering the idle timeout to 60s exposed a latent flaw: the read deadline is set **once per messageLoop iteration** as an absolute `T0 + IdleTimeout`, but `COPY FROM STDIN` reads its entire `CopyData` stream in a loop *within one iteration* and never reset it. At 24h this never bit; at **60s a COPY streaming longer than the timeout dies mid-stream** — precisely the COPY-heavy `posthog_data_import*` importers we're trying to help.

**Fix:** `armIdleReadDeadline()` is now called before **each** in-loop client read (all four COPY variants + the shared messageLoop), so the timeout measures **inter-message idle** — an actively-streaming COPY never times out, a stalled/abandoned one is still reaped. Added `copy_active_and_survives_idle` (streams a COPY over ~66s, asserts it survives with all rows).

### ✅ Verified clean (review's other findings)
- **No leak to workers** — the 60s applies only to the CP's client-facing PG loop; `ProcessIsolation` is force-off outside standalone, the k8s worker-pod idle timeout is a *different* field (5m), and duckdb-service workers serve Flight, not this loop.
- **`NormalizeIdleTimeout`** — standalone refactor is behavior-identical; `-1s` genuinely disables; no double-normalization.
- **Long ACTIVE queries not killed** — the loop is in `handleQuery` during execution, not reading; the disconnect monitor clears its deadline on stop.
- **Flight ingress** unaffected.

### On your two extra asks
- **COPY categorized correctly in Live** — yes: `handleQuery` stores `currentQuery`+`queryStart` *before* COPY dispatch, so `connState()` is `active` and `elapsed_ms` ticks the whole stream (won't be flagged idle by the #853 indicator). The harness test asserts this.
- **COPY path is fast** — the remote/warehouse case uses the **streaming path** (`handleCopyInRemoteStreaming` → DoPut to the worker → DuckDB's CSV loader), no CP-side buffering; `copyDataWireReader` is a clean single-copy `io.Reader`. The full-buffer paths are only the BLOB-column / binary-format edge cases. No perf change needed.

PR #855, opened 2026-07-01
admin/live: cancel sessions by worker id (fix cross-CP cancel)

fuziontech:
## The bug

The Live view's **Cancel** button fails with `{"error":"no active session with pid 1176"}`:

- `KillSession(pid)` scans only the **serving replica's** stacks, but behind the admin ALB (6 CP replicas) the session almost always lives on a *different* CP than the one the cancel request landed on → not found.
- And we **can't** just fan the pid-based cancel out: post-#851 pids are unique *within* a CP but still **collide across CPs** (each starts at 1000), so a pid fan-out could kill the *wrong* replica's session.

This is the same class of issue the detail view already solved by addressing on the cluster-unique worker id.

## The fix

Cancel by **worker id** (cluster-unique, config-store issued) with fan-out — collision-safe:

- `KillSessionByWorkerID(wid)` destroys the session on that worker on this replica (0/1), via `SessionForWorker`.
- `POST /api/v1/sessions/by-worker/:wid/cancel`: kills locally, and **only if this replica didn't own it** fans out to peers (`?scope=local` recursion guard, `PostPeers` + `sumKilled` — same plumbing as the per-user kill switch); `404` only if no replica owns the worker. The old `/sessions/:pid/cancel` route stays for direct API use (documented local-only).
- **UI**: the query-row, session-row, and detail-dialog Cancel buttons all address by `worker_id`.

## Tests
- **`TestCancelByWorkerFansOut`** — local hit skips the fan-out; peer-owned resolves via fan-out; `scope=local` never recurses; unknown worker → 404.
- **`harness.sh admin_cancel_by_worker`** — holds a real session, cancels it by worker id (`killed>=1`, session disappears), and asserts an unknown worker → 404.

Note: the idle sessions you were trying to cancel will also now self-evict via the 60s idle timeout (#854) once it's fully rolled — this makes the *manual* cancel button work regardless of which replica serves it.

### DIFF
--- a/controlplane/admin/README.md
+++ b/controlplane/admin/README.md
@@ -64,7 +64,8 @@ Added for the console:
 | `GET /api/v1/sessions`, `/workers` | viewer | live sessions / session-holding workers |
 | `GET /api/v1/workers/fleet` | viewer | cluster worker counts by lifecycle state |
 | `GET /api/v1/cluster/instances` | viewer | live CP replicas (self-flagged) |
-| `POST /api/v1/sessions/:pid/cancel` | admin | tear down a session + its worker |
+| `POST /api/v1/sessions/:pid/cancel` | admin | tear down a session by pid — LOCAL only (pid is per-CP); prefer the worker-id form |
+| `POST /api/v1/sessions/by-worker/:wid/cancel` | admin | tear down the session on a cluster-unique worker id; fans out to whichever CP owns it (pid can't be fanned out — it collides across CPs). Returns `{killed, cp_responders, cp_total}` |
 | `POST /api/v1/orgs/:id/users/:username/kill` | admin | per-user kill switch (one-shot): tear down ALL of a user's sessions + in-flight queries cluster-wide. Returns `{killed, cp_responders, cp_total}`. Does NOT block reconnects |
 | `POST /api/v1/orgs/:id/users/:username/disable` | admin | persist `disabled=true` (refused at connect on PG wire + Flight), reload the snapshot cluster-wide so the block is immediate, AND kill the user's live sessions. Returns `{disabled, killed, …}` |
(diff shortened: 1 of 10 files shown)


fuziontech:
## Adversarial review — no blockers; clean

Independent deep review + my own pass. **No blockers, no should-fix. No human reviews on the PR** (only the test-impact bot). CI is fully green on the commit (all 9 jobs, incl. every integration job).

### Verified
- **Collision safety (the crux) — correct.** Worker id is genuinely cluster-unique in the multi-replica remote topology: it's allocated by a globally-atomic Postgres sequence (`CreateSpawningWorkerSlot` → `nextval(worker_records_id_seq)`, seeded once above `MAX` and never re-seeded backward). So exactly one CP owns the session; the fan-out can't double-kill or hit the wrong one, and the `killed==0` guard never needlessly fans out on a local hit.
- **`DestroySession` — identical teardown to the old pid path**: closes the executor, issues the worker-side destroy RPC, releases the worker to hot-idle. **Active in-flight queries are torn down**, exactly as the old button did for same-CP sessions — only the *addressing* changed.
- **Auth/audit** — same group as the other kill routes (RoleGate admin-only + audit); peer fan-out authenticates with the internal secret; scope=local peer legs are audit-skipped (one row per operator action).
- **UI** — `RunningQuery`/`SessionStatus`/`QueryDetail` all carry `worker_id`; all three Cancel buttons + the api client use it; no pid left anywhere.
- **`?scope=local` recursion guard** honored; 404 path correct.

### Nits (both addressed in the follow-up commit)
1. `cp_responders`/`cp_total` misled on this single-owner op (non-owning peers 404 → undercount) — dropped them; the response is now just `{killed}`.
2. Bumped the harness session-hold above its appear-poll budget to kill a theoretical slow-CI false-negative.

PR #860, opened 2026-07-01
docs(design): billing pull API (buckets by org/team/cpu/mem)

benben:
## What

Draft spec (`docs/design/billing-pull-api.md`) for the **pull-based** billing integration the billing team wants: billing pulls compute-usage buckets from a small duckgres HTTP API instead of duckgres pushing capture events.

## Decisions baked in

- **Buckets keyed by** `(org_id, team_id, cpu_millicores, mem_mib, bucket_start[60s])`; values `cpu_seconds` (vCPU-s) + `memory_seconds` (GiB-s).
- **API:** `GET /api/billing/buckets?limit=N` (closed, un-acked, oldest first) + `POST /api/billing/buckets/ack {ids}` (idempotent delete). Bearer auth over the existing posthog→mw ingress.
- **No data loss:** delete only on explicit ack; at-least-once + billing dedups by natural key; only closed buckets served.
- **No infinite accumulation:** ack deletes immediately; **30-day safety GC** (logged) bounds the table.
- **team_id:** config-store org→team lookup at connect (needs plumbing).
- **Replaces** the capture push (removes ingest URL/token + drain-to-PostHog); keeps per-connection metering.

Draft for review — implementation follows once the shape is agreed.

### DIFF
--- a/docs/design/billing-pull-api.md
+++ b/docs/design/billing-pull-api.md
@@ -33,9 +33,12 @@ Internally, one row per unique key, values accumulated:
   same bucket and fractional sizes keep full precision.
 - **Values:** `cpu_seconds` (vCPU-seconds), `memory_seconds` (GiB-seconds) —
   summed over every connection that falls in that key + minute.
-- `team_id` (config-store **org → team** lookup) and `query_source` (the worker's
-  role: standard query vs endpoints) are resolved at connection time — both need to
-  be threaded onto the connection.
+- `team_id` is the org's **default team** for now — a fixed value per org (from the
+  config-store org→team default), reported so the shape is right, but **no per-team
+  attribution logic** yet. A "team" is really a schema and one connection can span
+  several, so true per-team split is future work; today every bucket carries the
+  org's default team. `query_source` (the worker's role: standard query vs
+  endpoints) is resolved at connection time and threaded onto the connection.
 - Worker size (`cpu`, `mem_gib`) is part of the key, so different worker sizes
   accumulate — and bill — separately. Units match the values (vCPU / GiB).
 
@@ -149,9 +152,9 @@ sizes; stored as `NUMERIC` so grouping is exact.)
 - **Extend:** the bucket table gains `team_id`, `query_source`, `cpu`, `mem_gib`
   (`NUMERIC`) in the key (new migration); add a single `last_acked` cursor row; add
   the HTTP API (aggregate-on-read into one row per key + watermark ack) + safety GC.
-- **Add:** config-store `org → team` lookup and the connection's `query_source`
-  (standard vs endpoints) — both threaded onto the connection; a bearer secret for
-  the API.
+- **Add:** the org's default `team_id` (fixed per org — no per-team logic yet) and
+  the connection's `query_source` (standard vs endpoints) threaded onto the
+  connection; a bearer secret for the API.
 
 ## Defaults / house-keeping
 
--- a/docs/design/billing-pull-api.md
+++ b/docs/design/billing-pull-api.md
@@ -2,27 +2,29 @@
 
 Status: **DRAFT**. Supersedes the reporting hop of
 [`billing-compute-seconds-plan.md`](./billing-compute-seconds-plan.md): billing
-now **pulls** buckets from duckgres instead of duckgres pushing capture events to
+now **pulls** usage from duckgres instead of duckgres pushing capture events to
 PostHog ingestion. The per-connection metering is unchanged — only how the data
 leaves duckgres changes.
 
 ## Overview
 
 duckgres meters compute usage into 60-second **buckets** in its config store. The
-billing service periodically **pulls** closed buckets over a small HTTP API,
-processes them, and **acks**; duckgres deletes acked buckets. A safety GC caps how
-long anything can linger. Direction is posthog → mw (billing initiates), which the
-existing duckgres control-plane ingress already supports.
+billing service periodically **pulls** the usage accumulated since its last ack,
+processes it, and **acks a watermark**; duckgres advances a cursor and deletes
+everything up to it. A safety GC caps how long anything can linger. Direction is
+posthog → mw (billing initiates), which the existing duckgres control-plane
+ingress already supports.
 
 ```
-duckgres:  connection ends → meter → accumulate into config-store buckets
-billing:   GET /buckets  → process → POST /buckets/ack  → duckgres deletes
-safety:    GC hard-deletes anything older than 30 days (logged)
+duckgres:  connection ends → meter → accumulate into 60s config-store buckets
+billing:   GET /usage  → process → POST /ack {watermark}
+                                     → duckgres advances cursor + deletes ≤ watermark
+safety:    GC hard-deletes buckets older than 30 days (logged)
 ```
 
 ## Buckets
 
-One row per unique key, values accumulated:
+Internally, one row per unique key, values accumulated:
 
 - **Key:** `(org_id, team_id, cpu_millicores, mem_mib, bucket_start)`
   where `bucket_start = floor(connection_end / 60s)`.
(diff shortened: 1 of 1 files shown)


docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,119 @@
...
+    "cpu_millicores": 8000,
+    "mem_mib": 16384,
+    "bucket_start": "2026-07-01T12:34:00Z",
+    "cpu_seconds": 80,
+    "memory_seconds": 160
+  }
+]
+```
+
+Billing keeps calling until it gets an empty list.
```
mcoll-posthog: This is somewhat scary to me, since buckets are 60 seconds long depending on latencies on usage reports etc we may be polling indefinitely and not be able to catch up.

I wonder if you could pre-agg and you give us daily team-level cumulatives since last ACK?

mcoll-posthog: Maybe you send us the last bucket that was aggregated so it's kind of a running pointer and we ACK that. We update the highwater mark basically.

mcoll-posthog (dismissed):

docs/design/billing-pull-api.md
benben: we can also move to hourly or daily buckets, I have no opinion here beside we'd loose granularity and wouldn't be able to show this to users.

>Maybe you send us 

I cannot send anything to you, there is only the direction you -> duckgres. duckgres cannot reach into prod VPCs (by design)

for the high watermark, how would that help the problem? you could still fall behind, no?

mcoll-posthog: yeah yeah, sorry bad wording. What I mean is that you return us that.

For the high watermark the idea is that you cannot really fall behind because you always get everything since last ACK until now instead of having to chase 60 second buckets forever.

mcoll-posthog: > we can also move to hourly or daily buckets, I have no opinion here beside we'd loose granularity and wouldn't be able to show this to users.

We only run usage reports daily right now (although this will improve soon). We'd store the deltas and accumulate them in our daily usage report anyway.

mcoll-posthog: I think it's good if you have smaller buckets, and then we aggregate. What I'm worried is that if usage reports for whatever reason takes 6 hours to start again that we need to catch up # customers x 360 buckets while it's growing every minute.

cohenroy: @mcoll-posthog i think i'm missing something, do you mean returning all the pre-aggregated buckets since the last one we acked? even with pre aggregation i we should have pagination
@benben while i agree that we lose granularity when we pre aggregate, i think that hourly buckets are good enough for this use case (I think daily will cause too much of a delay in showing the usage)

mcoll-posthog: No, hourly is not good enough for limiting, let's keep the small granularity and we'll query and aggregate usage with the highwater mark ACK system.

mcoll-posthog: > i think i'm missing something, do you mean returning all the pre-aggregated buckets since the last one we acked? even with pre aggregation i we should have pagination

we may think of a pagination system yes, but the important thing is that we get one bucket per customer dimension I think so the time to catch up doesn't depend linearly on the time passed since the last one.

benben: pushed some updates, please re-read

docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,135 @@
...
+- **One row per key regardless of elapsed time.** A pull returns everything since
+  the last ack aggregated, so the response size is bounded by the number of active
+  keys — **not** by how long billing was away. A week of downtime returns the same
+  row count, just larger sums. Billing can never fall behind chasing 60s buckets.
+
+### `POST /api/billing/ack`
+
+```json
+{ "watermark": "2026-07-01T12:40:00Z" }
+```
```
mcoll-posthog, line 115: This works if you return ALL buckets on a single call, right? if we have pagination on top this would consume stuff we wouldn't want. Should we just return all aggregated buckets each call?

benben: internally we have hour 60s buckets. when doing `GET /api/billing/usage` you will get all un-acked buckets since the last acked watermark at once, so this equals all data not yet in billing.

`POST /api/billing/ack` just acks these and duckgres would bump the watermark. It doesn't return anything here.

good?

mcoll-posthog: cool, so you return everything. Sounds good to me for now!

mcoll-posthog: Maybe I'm explaining myself very badly, when I said it "returns" I meant that `GET /api/billing/usage` would need to return all aggregated buckets, if at some point we could not return all for whatever reason (pagination, etc) then this shape wouldn't work that well. But if we return everything then it's good enough for me.

docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,142 @@
...
+    {
+      "org_id": "org_abc",
+      "team_id": "12345",
+      "cpu": 8,
+      "mem_gib": 16,
+      "cpu_seconds": 4800,
+      "memory_seconds": 9600
+    }
+  ]
+}
```
mcoll-posthog, line 88: Could we return the lower watermark together with the high watermark? For validation purposes so we can raise an issue if for whatever reason we missed a chunk on our side.

docs/design/billing-pull-api.md
benben: haha good old semantics, always hard :D 

sounds good to me, since we return only one row per bucket included the whole aggregate this only becomes a problem if we have SO MANY buckets that we kinda exceed repsonse size or timing, I think we are good here for a long time.

docs/design/billing-pull-api.md
benben: this is the last watermark you ack'ed that you got all buckets. so if anything goes wrong inbetween you simply wouldn't bump it higher or would reread already read buckets again and then ack. good?

docs/design/billing-pull-api.md
cohenroy: over time and adoption the response will get bigger (size wise) as we'll have more orgs/teams/distribution of workers, but i'm not sure if this is a real concern, wdyt?
@mcoll-posthog when you say "all aggregated buckets" you mean in a single value right? if that's the case, could we run into a scenario where different days usage is aggregated into a single response that we associate with the wrong day?

docs/design/billing-pull-api.md
mcoll-posthog: yeah, it's just for cross validation purposes, let's say that we have a bug and we ackd but for whatever reason we fucked up and it didn't work.

With the lower and higher we can add a small assertion to get more visibility and know that we don't have any holes.

mcoll-posthog: Without the low watermark we have no easy way of detecting that we fucked up under those circumstances (HTTP and postgres transactions are not atomic for example although we have the outbox system and temporal workflows that should mitigate that bringin it's own set of problems, but you get the idea).

docs/design/billing-pull-api.md
mcoll-posthog: yes, one of the dimensions should be the date so we get two different buckets during the day transition

docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,142 @@
...
+
+Returns compute usage **aggregated since the last ack**, one row per worker size
+per `(org_id, team_id)`. Size is reported as `cpu` (vCPU) and `mem_gib` (GiB) —
+same units as the values:
+
+```json
+{
+  "watermark": "2026-07-01T12:40:00Z",
+  "usage": [
+    {
```
cohenroy, line 77: @benben can we add a `endpoints` flag/tag?
```suggestion
    {
      "compute_type": "endpoints" | "standard",
```

docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,143 @@
...
+
+## API
+
+Control-plane HTTP, over the existing posthog → mw duckgres ingress.
+Auth: `Authorization: Bearer <internal secret>`.
+
+### `GET /api/billing/usage`
+
+Returns compute usage **aggregated since the last ack**, one row per worker size
+per `(org_id, team_id)`. Size is reported as `cpu` (vCPU) and `mem_gib` (GiB) —
```
mcoll-posthog: I think it should be

```suggestion
per `(org_id, team_id, date, cpu, mem_gib)`. Size is reported as `cpu` (vCPU) and `mem_gib` (GiB) —
```

docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,144 @@
...
+same units as the values:
+
+```json
+{
+  "watermark": "2026-07-01T12:40:00Z",
+  "usage": [
+    {
+      "compute_type": "endpoints" | "standard",
+    compute_type: "endpoints",
```
cohenroy:
```suggestion
    {
      "compute_type": "endpoints" | "standard",
```

docs/design/billing-pull-api.md
mcoll-posthog: or adding `compute_type` as well

docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,143 @@
...
+  keys — **not** by how long billing was away. A week of downtime returns the same
+  row count, just larger sums. Billing can never fall behind chasing 60s buckets.
+
+### `POST /api/billing/ack`
+
+```json
+{ "watermark": "2026-07-01T12:40:00Z" }
+```
+
+- duckgres advances its cursor to `watermark` and **deletes all buckets ≤
```
mcoll-posthog: Could this have a TOCTOU issue? I'm guessing the bucket insertions are not atomic through all customers. 

If that's the case then there is the following problem:

* We query when we have a partial insert of the current buckets at watermark T
* We process those buckets (for 1/2 our customers we process the buckets up to T-1)
* We send the ACK for T
* The system deletes all buckets where the time is <= T (we've missed 1/2 of the buckets)

mcoll-posthog: This will depend on implementation but it's something we need to take into account or we may lose usage data.

cohenroy: Isn't it handled by:
> A bucket is **closed** (eligible to serve) once `now ≥ bucket_start + 60s + grace`

mcoll-posthog: depends on the grace, and how long it takes to insert everything

mcoll-posthog: But yes, I guess it mostly covers it, I hadn't seen that (30s grace).

cohenroy: it should cover it for most of the cases (we can use a `60s` grace to be on the safe side), but it got me wondering if there's a real case where the insert is stuck for a long while (bug/incident) and we want to amend, but it's probably over engineering for now

docs/design/billing-pull-api.md
```diff
@@ -0,0 +1,171 @@
...
+  several, so true per-team split is future work; today every bucket carries the
+  org's default team. `query_source` (`standard` | `endpoints`) is set by a session
+  GUC (`duckgres.query_source`), defaulting to `standard` when unset; the meter
+  reads it per connection. (If it's changed mid-connection the per-connection meter
+  uses a single value — same future refinement as team.)
+- Worker size (`cpu`, `mem_gib`) is part of the key, so different worker sizes
+  accumulate — and bill — separately. Units match the values (vCPU / GiB).
+
+**60-second resolution is kept internally** (source of truth), but the API
+**aggregates on read**: a `GET` sums all closed buckets in the window into **one
```
cohenroy, line 48: One small note here, we need to make sure that when the low/high watermark are cross dates (like `now = 00:05` and last ack was at `23:50`), we should split it into different buckets so we can differentiate different day's usage

cohenroy (approved):
Left a small comments but LGTM

docs/design/billing-pull-api.md
benben: added a date key

PR #861, opened 2026-07-01
admin: Errors page — recent redacted query errors, cross-CP

fuziontech:
## What

Adds an **Errors** page to the admin console — a triage surface for failures happening across the cluster. The Live page shows what's *running*; this shows what's *breaking*.

Every failed query is captured into a bounded in-memory ring on each control plane at the single `logQueryError` tap, surfaced at `GET /api/v1/errors` and **merged/fanned-out across CP replicas** (an error belongs to exactly one CP, so the union is disjoint — concat, newest-first, capped). Filterable by org / user / SQLSTATE / category, with a click-through detail dialog carrying the full redacted context.

## The page

- **Table**: time (age + hover-for-timestamp), org, user, SQLSTATE, category badge (system → destructive, DuckLake conflict/metadata → warning, user → muted), truncated message, worker.
- **Detail dialog** (row click): time, org, user, pid, worker, worker pod, client addr, trace id, redacted message, redacted query.
- **Summary strip**: total recent errors vs system-attributable (page-worthy) count.
- **Filters** applied server-side after the cross-CP merge; polls every 5s.

## Redaction is load-bearing

The ring stores **only** the redacted forms already computed by `logQueryError` — `Query` via `RedactForLog`, `Message` via `RedactErrorForLog`. A failed `CREATE SECRET` echoes the offending SQL in its engine error, so both sinks are scrubbed at capture — the credential can never reach the Errors page. Covered by a unit test and an e2e assertion.

## Not durable history

This is a **live-triage buffer** (`DefaultRecentErrorCap=500` per server), independent of the query-log sink config. Long-term error history stays in the external query-log (Kafka) pipeline — see the doc note in `CLAUDE.md`.

## Layout

- **Backend**: `server/recent_errors.go` (ring + `Server.RecentErrors`), capture in `server/conn.go` `logQueryError`, `/errors` handler + fan-out in `controlplane/admin/live.go`, `clusterInfoProvider.RecentErrors`.
- **Frontend**: `pages/Errors.tsx`, `lib/errors.ts` (pure category/SQLSTATE helpers), `useErrors` hook, `api.listErrors`, `ErrorEntry` type, nav + route.

## Tests

- `server/recent_errors_test.go` — ring newest-first/eviction/nil-safety + the **redaction invariant** (a `CREATE SECRET` error must not leak its credential into the ring) + plain-error verbatim capture.
- `controlplane/admin/live_aggregate_test.go` — `/errors` cross-CP merge, newest-first sort, post-merge filter, `scope=local` fan-out guard.
- `controlplane/admin/ui/src/lib/errors.test.ts` — category variant/label, system classification, SQLSTATE bucketing (vitest).
- `tests/e2e-mw-dev/harness.sh::admin_recent_errors` — a marker query error is captured in `/errors` with its identity + SQLSTATE, and a failing `CREATE SECRET`'s credential is absent from the captured entry.


fuziontech:
## Self-review

Went through the diff end-to-end. One real bug found and fixed before it could land:

**🐞 Cross-CP fan-out storm (fixed).** The `/errors` handler originally fanned out to peers with the path `/api/v1/errors?limit=200`. But `clusterPeerFetcher.FetchPeers` builds the peer URL as `…%s?scope=local` — appending `?scope=local` to the path. With a `?limit=` already present, the peer URL became `…/errors?limit=200?scope=local`, where `scope=local` is **not a parsed query param** (it's swallowed into the `limit` value). So peers would not see the recursion guard → each peer re-fans-out to all peers → a cross-CP fan-out storm bounded only by the HTTP timeout.

Fix: the fan-out path is now **bare** (`/api/v1/errors`), matching every other aggregated endpoint. Peers return their whole local ring (a `scope=local` slice is returned un-filtered/un-capped), and the requesting CP owns the merge → sort → filter → limit. Added a regression assertion that the fanned path is bare and that the merge still yields local+peer.

**Checked and OK:**
- **Redaction** (load-bearing): the ring stores only `RedactForLog(query)` / `RedactErrorForLog(query, msg)`, computed once in `logQueryError`. Unit test + e2e both assert a failing `CREATE SECRET`'s credential never reaches the ring. e2e run showed `admin errors: redaction holds (credential absent…)` ✅
- **Nil-safety**: `recordRecentError` / ring `add`/`snapshot` are nil-receiver safe; the existing `logQueryError` tests (nil server) still pass.
- **Merge correctness**: errors are disjoint per CP (each captured on the owning connection's CP) → concat + newest-first, no worker-id dedup needed.
- **Frontend**: redacted `message`/`query` rendered as escaped text in `<pre>` (no XSS); `useErrors` query key is a memoized filter object (stable across renders).

**CI:** the two red checks on the first push were both pre-existing flakes, not this change — `TestOrgRouter…SharedWorkerActivator` (timing; passes 10× locally, doesn't touch this code) and `copy_active_and_survives_idle` (there's an active `ben/e2e-copy-idle-flake-hardening` branch for exactly this test). `controlplane-k8s-tests` already went green on re-run. Fresh CI + e2e running on the amended commit.

PR #875, opened 2026-07-02
Remove Iceberg/Lakekeeper support

benben:
## Summary

DuckLake is now the only catalog. This removes the per-org Lakekeeper Iceberg REST catalog end to end — the catalog attach/refresh path, the Lakekeeper provisioner, the iceberg transpiler backend, the config surface, the schema columns — plus all logic that existed only to support it.

132 files, +933 / −12,273.

### Removed

**Packages**
- `server/iceberg`, `server/icebergmeta`, `server/lakekeeperbroker`, `server/pgtypes`
- `controlplane/provisioner`: lakekeeper provisioner, client, k8s glue, postgres admin helpers
- `controlplane/lakekeeper_inputs.go`, `server/iceberg_drop_schema.go`
- `k8s/lakekeeper-networkpolicy.yaml`, `scripts/copy-ducklake-to-iceberg.sh`, the lakekeeper runbook

**Server**
- `AttachIcebergCatalog` / `RefreshIcebergSecret` / `setIcebergDefault`; the iceberg-only activation branch — `ActivateDBConnection` now requires a DuckLake `metadata_store`
- `Config.Iceberg` / `server.IcebergConfig`
- On-demand iceberg column-metadata loading (the `*WithMetadata` query wrappers are inlined to direct executor calls)
- The iceberg schema-evolution error classifier and `friendlyExecError` (call sites use `err.Error()`)
- The `LOAD iceberg` INSTALL special case; iceberg compat-view SQL in `catalog.go`/`sessionmeta` (incl. `PrimeIcebergCatalog` and the catalog-settle retry)
- `iceberg_sigv4`/`iceberg_oauth` dropped from the reserved secret names

**Control plane**
- Catalog selection is ducklake-only: a startup `database=iceberg` is now rejected with 3D000
- Per-user `default_catalog` selector removed end to end (model, snapshot, admin API, admin UI)
- `IcebergConfigForOrg`, the `Iceberg` field on activation payloads, Duckling CR iceberg spec/drift/teardown
- Provisioning API now requires `ducklake.enabled=true` (a warehouse without a catalog has nothing to attach)

**Config surface**
- `--iceberg-*` flags, `DUCKGRES_ICEBERG_*` env vars, the YAML `iceberg:` block
- The worker binary's Lakekeeper OIDC token broker (`DUCKGRES_LAKEKEEPER_TOKEN_PATH`)

**Transpiler**
- The iceberg backend profile, and the warn/error hybrid DDL guards (`WarnOnStrippedConstraints`/`ErrorOnSilentNullDefaults`) that only it enabled

**Schema**
- New goose migration `000013_drop_iceberg_columns.sql`: drops all `iceberg_*` columns from `duckgres_managed_warehouses` and `default_catalog` from `duckgres_org_users` (down migration restores them)

**Build / docs / tests**
- Dockerfiles no longer fetch/bundle the iceberg DuckDB extension
- Integration compose stack drops the lakekeeper services; e2e `run.sh`/manifests cleaned (the harness was already ducklake-only)
- CLAUDE.md / README / design docs updated in the same PR per the docs-in-sync rule
- Regression assertions added where behavior flipped: `database=iceberg` rejected, legacy `default_catalog`/`iceberg` request bodies ignored or rejected, `iceberg_sigv4` no longer reserved

### Not in this repo (follow-ups elsewhere)
- Crossplane Duckling XRD/composition: `spec.iceberg` block, the cnpg-shard iceberg CEL rule, per-org Lakekeeper composition resources
- Charts: lakekeeper SA-token projected volume / networkpolicy templates
- Legacy Duckling CRs still carrying `spec.iceberg` are ignored by the CP

## Test plan

- `go build` + `go vet`, with and without `-tags kubernetes`: clean
- Full `go test ./...` (both tag sets): green locally except Postgres-container suites (no docker on the dev box; CI runs them) and four `tests/integration` tests that fail identically on unmodified `main`
- Migration test updated to assert goose v13 and the dropped columns (runs in CI against real Postgres)
- Admin UI `vite build`: clean
- e2e-mw-dev: per-PR workflow exercises the full ducklake activation pipeline on this PR


benben:
## Adversarial review (fresh-context agent) — outcome

A dedicated adversarial reviewer diffed every hunk against `origin/main`, tried to falsify the change, and ran both tag sets' builds/vet/tests. **Verdict: merge-safe for all in-repo code paths** — every shared-path hunk (session init, catalog resolution, USE rewriting, error classification, query wrappers, extension loading, compat views, migration 000013, wire payload compat) reduces to byte-identical behavior for DuckLake/memory/standalone sessions. Notable refuted candidates: extended-query ctx regression, DROP COLUMN misclassification, `friendlyExecError` semantics, migration column-set mismatch, mixed-version activation payload decode.

### Fixed in f8afb86
- **Dead warnings plumbing removed**: `Result.Warnings` had zero producers after the iceberg DDL guards went away — field, copy sites, and NoticeResponse loops deleted.
- **Unused `ctx` param dropped** from `execCompatibilityFallback`.
- **Rolling-deploy cosmetic leak plugged**: `__duckgres_iceberg_column_metadata` stays in the compat-view exclusion lists (with a "legacy, remove later" comment) so hot-idle workers warmed by pre-removal binaries don't surface it as a user table mid-deploy.

### Deploy gates (operational, not code — sequence with the merge)
1. **Existing per-org Lakekeeper resources are now orphaned on deprovision.** The teardown path (`DeleteForOrg`: Lakekeeper CR/SA/Secret + per-org Postgres DB/role on the cnpg shard) is gone, and deprovision reports success while leaving them behind. Needs a one-off sweep decommissioning all live per-org Lakekeeper instances/DBs, and existing Duckling CRs with `spec.iceberg.enabled=true` should be patched (the CP no longer drift-corrects them, so the composition may keep reconciling Lakekeeper for those CRs).
2. **Crossplane XRD ordering**: the old code asserted cnpg-shard ducklings *must* have `iceberg.enabled=true` (XRD CEL rule, other repo). New CRs no longer carry the block — if the rule still exists, cnpg-shard provisioning gets admission-rejected until the XRD ships. The per-PR e2e-mw-dev run provisions a cnpg-shard duckling, so a green run on this PR disproves the hazard — confirm before merging.
3. **Orgs with `ducklake_enabled=false`** (previously iceberg-only) now fail activation with a clear error ("tenant activation requires a ducklake metadata_store"). Confirm no live org is in that state, or accept the outcome.
4. Minor mixed-fleet note: admin `PUT .../warehouse` strict-decodes and now 400s on bodies carrying `iceberg`/`iceberg_state` (deliberate, regression-tested) — any external tooling doing GET-modify-PUT against an old-CP response needs to drop those keys. The provisioning API is lenient and just ignores a stale `iceberg` field.

fuziontech (approved):

PR #907, opened 2026-07-07
Add Slack org lifecycle notifications

fuziontech:
## Summary

Adds optional Slack operational notifications for org and managed-warehouse lifecycle events.

- Adds an async notification subsystem with a bounded queue, delivery timeout, Slack incoming-webhook sink, and Prometheus counters for queue drops/delivery failures.
- Emits `org_created` only when an org row is newly inserted, `org_deleted` on successful admin deletion, plus warehouse provision/deprovision lifecycle notifications.
- Keeps Slack payloads conservative for a public repo: raw org IDs, database names, endpoints, secrets, buckets, SQL, and customer names are omitted; messages use only an allowlist of low-risk fields by default. Operators can set env-only `DUCKGRES_NOTIFICATION_HASH_KEY` to include a stable HMAC org reference for Slack-side correlation.
- Documents env-only Slack configuration, defaults, optional HMAC correlation, failure recovery, notification health metrics, and the intentional no-drain shutdown behavior.

## Claude review follow-up

Ran Claude Code review with Sonnet. It found no blocking correctness or data-leak bugs. I addressed the non-blocking gaps it raised:

- Added terminal provisioner notification assertions alongside analytics tests.
- Added `org_deleted` notifications for successful admin deletes.
- Added `duckgres_notifications_dropped_total` and `duckgres_notification_delivery_failures_total`.
- Documented that shutdown does not drain the notification queue because delivery is best-effort.

## Final self-review follow-up

- Replaced the unsalted SHA-256 org reference with an optional HMAC org reference gated by `DUCKGRES_NOTIFICATION_HASH_KEY`.
- Default Slack payloads no longer include any org correlation token unless the secret hash key is configured.

## Validation

- `go test ./internal/notifications ./controlplane/provisioning`
- `go test ./internal/notifications ./internal/cliboot ./controlplane/provisioning`
- `go test -tags kubernetes ./controlplane/provisioner`
- `go test -tags kubernetes ./controlplane/admin -run 'TestCreateOrg|TestRegisterAPI|TestCreateUser|TestListOrgs'`
- `go test -tags kubernetes ./controlplane/admin -run 'TestDeleteOrgRejectsWhenWarehouseStillExists|TestDeleteOrgSucceedsAfterWarehouseRemoved|TestDeleteOrgCascadesDeletedWarehouse$'`
- `just test-unit`
- `just lint`

Note: `go test -tags kubernetes ./controlplane/admin` still reaches existing Postgres-container tests and fails locally because the container cannot start in this environment; narrowed admin handler tests pass. `just lint` initially exposed a local tooling issue because the installed `golangci-lint` binary was built with Go 1.25 and panicked on Go 1.26 files. I rebuilt the same `golangci-lint` v2.8.0 with local Go 1.26.4, then lint passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,10 @@ Run with config file:
 | `POSTHOG_HOST` | PostHog ingest host | `us.i.posthog.com` |
 | `ADDITIONAL_POSTHOG_API_KEYS` | **(Experimental)** Comma-separated list of additional PostHog API keys to publish logs to. Requires `POSTHOG_API_KEY` to be set. | - |
 | `DUCKGRES_IDENTIFIER` | Suffix appended to the OTel `service.name` in PostHog logs (e.g., `duckgres-acme`); only used when `POSTHOG_API_KEY` is set | - |
+| `DUCKGRES_SLACK_WEBHOOK_URL` | Slack incoming webhook URL for operational notifications. Env-only because it is secret-bearing. | - |
+| `DUCKGRES_NOTIFICATION_HASH_KEY` | Optional secret key used to include a stable HMAC org reference in Slack notifications. Env-only because it is secret-bearing. | - |
+| `DUCKGRES_NOTIFICATION_QUEUE_SIZE` | In-process notification queue size before events are dropped instead of blocking user requests | `100` |
+| `DUCKGRES_NOTIFICATION_TIMEOUT` | Per-notification delivery timeout | `2s` |
 
 ### PostHog Logging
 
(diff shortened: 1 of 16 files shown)


bill-ph (approved):
Reviewed current head 87621de. No P0 blockers found.\n\nThe notification queue, timeout, payload allowlist/HMAC behavior, lifecycle emission, and failure metrics are covered by tests; CI and security checks are green. Non-blocking operational note: shutdown intentionally drops queued notifications and queue saturation drops events, so alerting remains best-effort as documented.\n\n— Robo Bill

PR #929, opened 2026-07-10
feat(billing): skip unresolved-team (team_id 0) usage in the pull

cohenroy:
Skips managed-warehouse usage for orgs with no resolved default team (`team_id 0`) so it never enters the billing pull. Such usage is unattributable — billing keys the usage mirror by team alone, and `0` collides across orgs.

- **Source**: compute meter + storage sampler skip `team_id <= 0` (counted via `duckgres_*_skipped_no_team_total` for alerting; should stay flat at zero post-backfill).
- **Serve**: `GET /billing/usage` drops `team_id <= 0` rows — covers any buckets written before this shipped.

**To confirm**: assumes there are no pre-existing `team_id 0` buckets that ought to be billed. The serve-side filter makes the change safe regardless of store contents, but flagging for anyone who knows whether teamless billable orgs exist today.

### DIFF
--- a/controlplane/compute_billing_api.go
+++ b/controlplane/compute_billing_api.go
@@ -93,12 +93,42 @@ func (h *billingAPIHandler) getUsage(c *gin.Context) {
 	if storageRows == nil {
 		storageRows = []configstore.StorageUsageRow{}
 	}
+	// Never serve usage for an unresolved default team (team_id 0): it can't be
+	// attributed to a billable team and would collide across orgs on billing's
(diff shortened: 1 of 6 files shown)


cohenroy:
@benben Sending data when we can't attribute to a team is redundant + creates problems downstream (theoretical cross org leak), as far as I understand it shouldn't really happen + if that's the case can we make sure we don't have such values in the DB for whenever I merge the consumer?

benben:
fixed in #934

PR #947, opened 2026-07-14
reshard: fail the orphan wait fast on RBAC Forbidden + log observed MR policies

benben:
## Context

The first live cnpg→external reshard on mw-dev failed at `orphaning_source`: the runner set `retainCnpgOnFlip=true` (the flag stuck — the XRD-compat read-back passed), then polled `CnpgSourceMRsOrphaned` for the full 15m flip timeout and gave up with the generic "the composition may not honor retainCnpgOnFlip; charts/composition skew?" error.

The op log shows the real cause on every single poll: reading the provider-sql `Database`/`Role` managed resources (`postgresql.sql.m.crossplane.io`, ducklings namespace) is **Forbidden** for the runner's ServiceAccount. The duckling-reader grant covers only `ducklings` — the orphan wait is the one reshard step that reads beyond the Duckling CR, and it never had the RBAC it needs. The composition and XRD were fine (the retain conditional was already in the composition revision serving the op), and the composition demonstrably re-rendered `managementPolicies` when the flag was cleared on rollback — the runner was simply blind.

## Changes

- **Fail fast on Forbidden**: an RBAC denial is permanent — polling to the timeout cannot heal it and ends in an error blaming the wrong component. The wait now aborts immediately with the exact missing grant named (get on roles+databases in `postgresql.sql.m.crossplane.io`), and rolls back cleanly (retain flag cleared, source untouched, org back to ready).
- **Observable wait**: `CnpgSourceMRsOrphaned` now returns a per-MR observation (found/not-found + current `managementPolicies`, with "absent = default full lifecycle" made explicit). The existing 15s periodic wait log carries it, the wait announces itself (flag confirmed + how long it will wait), and the timeout error includes the **last observation** — a stuck wait is diagnosable from the op log alone.
- Docs: `docs/design/resharding.md` + `CLAUDE.md` no longer claim the runner needs "no new grants"; the RBAC requirement and the fail-fast semantics are pinned.

## Tests

- `TestReshardCnpgToExtOrphanWaitForbiddenFailsFast` — a Forbidden read (wrapped exactly like the real client wraps it, proving `IsForbidden` sees through the `%w` chain) fails the op immediately (1h flip timeout in the test — the 30s harness guard would trip if it polled), names the grant, never flips, never drops the source, resets the retain flag, unblocks the org.
- `TestReshardCnpgToExtOrphanWaitTimeoutReportsObservation` — a composition that never re-renders no-Delete times out with the last observed `managementPolicies ["*"]` in the error.
- `TestDescribeManagementPolicies` — pins the op-log rendering and its agreement with the conservative `managementPoliciesHaveDelete`.

`go build ./...`, `go build -tags kubernetes ./...`, `go vet -tags kubernetes ./...`, `go test -tags kubernetes ./controlplane/provisioner/` all green. The cnpg→ext positive path remains unit-only in the e2e harness (documented limitation: the harness lacks the RDS password), so no harness change.

## Not in this PR (charts repo)

The actual RBAC grant is a duckgres **chart** fix — see the PR comment below for the exact operator action.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -582,7 +582,11 @@ entrypoint), `controlplane/reshard_pod.go` (spawner) +
 - **cnpg→ext escape hatch = orphan-adopt then verified-delete** (charts
   `retainCnpgOnFlip`): copy → verify source → set `retainCnpgOnFlip=true` AND
   poll the cnpg Role/Database MRs until they carry the no-Delete policy (two-step
-  flip, closes the un-render-before-policy race) → flip type to external (now
+  flip, closes the un-render-before-policy race; this MR read needs an explicit
+  get grant on roles+databases in `postgresql.sql.m.crossplane.io` — NOT covered
+  by duckling-reader; a Forbidden fails the wait immediately with the missing
+  grant named, and the periodic wait log + timeout error carry the observed
+  `managementPolicies`) → flip type to external (now
   ORPHANS, not deletes, the cnpg role/DB) → verify the external catalog row
   counts match the copy EXACTLY → only THEN `DROP DATABASE` the retained source +
   clear the flag. ANY failure before that drop → flip back to cnpg-shard +
--- a/controlplane/provisioner/k8s_client.go
+++ b/controlplane/provisioner/k8s_client.go
@@ -843,12 +843,23 @@ const cnpgTenantMRGroup = "postgresql.sql.m.crossplane.io"
 // source's). The reshard runner polls this after setting retainCnpgOnFlip=true
 // and only flips once both MRs reflect the policy — closing the race where the
 // flip un-renders the MRs before the retain policy propagated.
-func (d *DucklingClient) CnpgSourceMRsOrphaned(ctx context.Context, name string) (orphaned, present bool, err error) {
+//
+// detail is a one-line human-readable account of what was observed (per-MR
+// found/not-found + current managementPolicies), destined for the reshard op
+// log so a wait that is not converging is diagnosable from the log alone.
+//
+// NOTE this read needs RBAC the Duckling patch itself does not: get on
+// roles+databases in cnpgTenantMRGroup (the duckling-reader grant only covers
+// ducklings). A Forbidden here is a permanent misconfiguration — callers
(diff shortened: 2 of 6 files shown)


benben:
**Root-cause record + required charts/operator action** (investigation of the failed live cnpg→ext reshard, mw-dev):

**What happened**: every poll of the 15m `orphaning_source` wait failed with
`... is forbidden: User "system:serviceaccount:duckgres:duckgres" cannot get resource "databases" in API group "postgresql.sql.m.crossplane.io" in the namespace "ducklings"` (same for `roles`). Confirmed live with `kubectl auth can-i` as that SA: `no` for both resources, `yes` for `ducklings`.

**What was ruled out**:
- *Composition/XRD skew (ArgoCD lag)*: the composition revision pinned by the Duckling during the op (created the evening before) already contained the `retainCnpgOnFlip` conditional; the XRD carries the field and the runner's read-back gate passed. The manual Argo sync produced a new revision only *after* the op had already failed — it is not what the op was missing.
- *Composition not re-rendering*: the composed-resource SSA manager re-applied `spec.managementPolicies` on the Role MR at the exact moment the rollback cleared the flag — the conditional works; the runner just couldn't read the result.
- *"Worked earlier the same day"*: the earlier successful ops were cnpg→cnpg shard moves, which never execute the orphan wait. This op was the first live exercise of the `orphaning_source` gate (the e2e covers cnpg→ext at unit level only).

**Required fix (charts repo, `charts/duckgres` RBAC)**: grant the control-plane ServiceAccount (which the reshard runner pod inherits) `get` on `roles` + `databases` in API group `postgresql.sql.m.crossplane.io`, namespace `ducklings` — e.g. a namespaced Role+RoleBinding in `ducklings` next to the existing duckling-reader binding (keep it separate from `duckgres-duckling-reader` per the existing escalation-prevention convention). Until that grant is deployed, any cnpg→ext reshard retry will fail (now immediately and with this exact instruction, thanks to this PR, instead of after a silent-looking 15 minutes).

**Retry readiness**: composition + XRD are ready; only the RBAC grant is missing.

PR #950, opened 2026-07-14
reshard: diagnose silent recovery waits — periodic observations + stranded-password classification

benben:
## Incident shape

During a cnpg→external reshard, the post-flip **recovery** (`recoverFromExternal`) hit its adopt-wait with a re-adopted role whose *actual* Postgres password differed from the freshly regenerated status password (a charts-side bug, fixed separately). Every probe failed with `FATAL: SASL authentication failed` — for ~8 minutes — and the op log showed **nothing** between the recovery announce and the eventual timeout. The operator saw a "stuck" reshard with zero diagnostics; the SASL failure was only found by reproducing the probe manually from a shell.

#947 fixed exactly this observability gap for the orphan-policy wait. This PR gives the remaining wait loops the same treatment.

## What each loop now logs

Every wait loop announces its **deadline** on entry, logs **what it observes** every ~15s (shared `progressLogInterval` knob, test-shrinkable; drain + orphan-wait moved onto it too), and its timeout error/log line carries the **last observation** so the terminal op-log line names the root cause:

- **`recoverFromExternal`** (the incident loop, cnpg→ext post-flip rollback): duckling read failures, endpoint/Ready state (+ `Synced=False` message), and classified probe errors; the timeout line ends with `last observation: …`.
- **`flipToCnpg` converge wait**: same treatment; the `did not become ready within …` error now carries the last observation (probe failures classified here too).
- **`flipToExternal` ESO/Ready wait**: already logged periodically — now the periodic line says what it observes (`duckling type/ESO password synced/ready` + `Synced=False`), plus deadline announce and last observation in the timeout error.
- The rollback flip-backs (cnpg→cnpg, ext→cnpg) are patch-and-move-on with no wait loop — nothing to instrument there.

## Stranded-password classification

`isAuthProbeError` recognizes authentication probe failures — SQLSTATE `28P01`/`28000` via the wrapped `*pgconn.PgError` chain (matching how `catalog_copy.go::Probe` wraps pgx connect errors), plus the `SASL authentication failed` / `password authentication failed` message texts for pre-flattened errors. `describeProbeFailure` then emits the operator-facing line:

> tenant-role probe failing with an AUTHENTICATION error: … — the role's actual Postgres password likely differs from the published status password (stranded password). If this persists, align it manually on the shard primary: `ALTER ROLE <role> WITH PASSWORD '<status password>'` (take the password from the duckling CR status — it is never logged here)

Deliberately **not** fail-fast: a composition fix can converge the password mid-wait, and bailing early would leave the org worse off — the loop polls to the deadline, but names the condition the whole way. No password can reach the op log: pgx `ConnectError`/`PgError` texts carry user/database only, and endpoints are logged via `Redacted()` (verified against pgx v5.9.2).

## Tests

- `TestProbeErrorAuthClassification`: classifier against realistically wrapped pgx errors (PgError-in-chain and flattened strings), plus negatives (connection refused, 57P03, ctx deadline).
- `TestReshardExtRecoveryProbeAuthFailureDiagnosed` (regression for the incident): recovery whose probe persistently fails SASL must produce the deadline announce, ≥2 periodic classified observations naming the stranded password + ALTER ROLE remedy, a timeout line carrying the root cause, no password anywhere in the log — and recovery still completes flip-back+adopt with the warehouse unblocked.
- Extended the existing flip-timeout tests to assert both timeout errors carry `last observation: …`.
- e2e (`tests/mw-dev/e2e/harness.sh`): the bogus-shard rollback lane now asserts the converge-wait deadline announce, periodic `waiting for target:` observations, and the last-observation-carrying flip-timeout error. (The recovery SASL path itself can't be provoked in-harness — it needs a role whose actual password diverged from the status password — noted in the harness comment; pinned by the unit regression instead.)
- Docs: `docs/design/resharding.md` gains a paragraph on the wait-loop diagnostics + the stranded-password diagnosis/remedy.

`go build ./…`, `go build -tags kubernetes ./…`, `go vet -tags kubernetes ./…`, `go test -tags kubernetes ./controlplane/provisioner/ -count=1` all green.

### DIFF
--- a/controlplane/provisioner/reshard_runner.go
+++ b/controlplane/provisioner/reshard_runner.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/posthog/duckgres/controlplane/configstore"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
@@ -61,6 +62,13 @@ type ReshardRunner struct {
 	// loopPoll is the inner-loop poll cadence (drain checks, flip waits,
 	// cancel checks). 5s in production; tests shrink it.
 	loopPoll time.Duration
+	// progressLogInterval is how often a wait loop (drain, flip converge,
+	// orphan-policy wait, recovery) writes what it currently observes to the
+	// op log. 15s in production; tests shrink it. Every wait loop MUST emit
+	// periodic observations at this cadence — a silent multi-minute poll loop
+	// is undebuggable from the op log (the mw-dev recovery that spun for ~8
+	// minutes on SASL auth failures with zero diagnostics).
+	progressLogInterval time.Duration
 
 	// extPasswords holds the EPHEMERAL external passwords by op id: the
 	// cnpg→ext target password (pulled from the creating CP replica's stash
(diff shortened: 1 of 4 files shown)


benben:
**e2e failure root-caused** (two consecutive `FAIL: reshard ext→cnpg: report missing` on the previous head — same assertion also hit other PRs' runs earlier today):

The op **succeeded** and **did** write its report — the runner pod's stdout in the run diagnostics shows `==== reshard report (succeeded) ====` seconds before the failing grep. The catalog copy processed **20,856 tables** (the shared ext-RDS catalog accumulates `ducklake_*` tables across e2e runs) and logs one op-log line per table, so the op log has ~21k entries. `reshard_log_has` fetched only a single first page (`after_id=0&limit=2000`, the server's per-page cap) and was blind to everything past entry 2000 — including the terminal report block. Growth is monotonic, which is why the lane went from sporadic to failing consistently today; this PR's extra wait-loop lines were incidental.

Fixed in 590353e: `reshard_log_fetch` pages the op log with `after_id` until exhausted; `reshard_log_has` greps the full log, `reshard_dump_log` now shows the true tail, `reshard_backup_debug` filters the full log. Pager verified against a faked multi-page endpoint (4005 entries / 3 pages, `set -eu`).

Side note for e2e infra (separate follow-up): the shared ext-RDS catalog's unbounded growth also inflates the ext→cnpg copy itself (~6 min, ~95MB, 856k rows this run) — worth a periodic cleanup, or the lane's budgets will eventually blow regardless of log paging.

benben:
**Status note for reviewers:** the red/cancelled `e2e` check on this head is expected — the rerun was cancelled deliberately, because each run against the shared ext-RDS database now drags a ~21k-table catalog through the reshard lane (the very failure mode diagnosed above). #952 (stacked on this branch) moves the e2e's ext org onto a per-run database, killing the unbounded catalog growth.

Validation order: e2e proves itself on #952 (small per-run catalog → fast lanes, and it exercises this PR's harness paging fix along the way, since #952 contains these commits). This PR then merges on the strength of that stacked green run + its unit tests, or gets one final quick e2e after the stack lands.

benben:
Superseded by #952, which carries this branch's commits (stacked) plus the per-run e2e ext database that fixes the slow/red e2e lane this PR was blocked on. Merging #952 delivers both.

PR #952, opened 2026-07-14
e2e: per-run external metadata database on the shared RDS

benben:
**Stacked on #950** (both touch `harness.sh`; rebase onto main if #950 merges first).

## Problem

The e2e's ext-backend org pointed every run at ONE shared database on the shared RDS, so its DuckLake catalog grew monotonically across runs — **20,856 `ducklake_*` tables** observed in today's runs. That inflated every catalog copy (the reshard ext→cnpg lane reached ~6 min / ~95MB / 856k rows for a run that creates a handful of tables — and it only gets worse), triggered the op-log-volume failure #950 diagnosed, and exposed runs to each other's writers (cross-run/foreign-writer interference class).

## Change

Each run provisions its ext org against a **run-unique database** `mdstore_e2e_<utc-ts>_pr<pr>` on the same RDS — same endpoint, same user, same SM secret (no SM/ESO/IAM changes; the provisioning API passes `metadata_store.external.database` through verbatim, verified in `controlplane/provisioning/api.go`).

- **Create** (`ext_rds_setup`, a lane concurrent with the readiness waits): the harness reads the ESO-synced RDS password from its **own** ext duckling's CR status (`status.metadataStore.password` — the harness SA already has cross-namespace duckling read RBAC), then `CREATE DATABASE`. No ordering hazard: the CP's end-to-end provisioning probe targets the per-run database and retries non-terminally (`controller.go reconcileProvisioning`), so `ready` flips right after the create lands.
- **Drop** (`ext_rds_teardown`, end of a passing run): `DROP DATABASE … WITH (FORCE)` + assert gone. Safe by then — `reshard_ext_to_cnpg` has already moved the org off the RDS, so the database is an orphaned source.
- **Stale GC** (start of every run, failure backstop): drops `mdstore_e2e_*` databases **older than 24h**, name-gated to the full per-run pattern with a parseable 14-digit UTC timestamp. A concurrent run's live database can never match (runs finish in <1h), and nothing outside `mdstore_e2e_*` is ever touched — the legacy shared database and any manually-created databases are structurally excluded. `run.sh` teardown / `e2e-cleanup` can't do this drop themselves (the GitHub runner has neither psql nor a path to the RDS password) — documented inline; scheduled main runs keep the GC cadence.

Also moves the kubectl bootstrap ahead of the readiness lanes (the extdb lane must not race the background download) and notes the new lane in the PASS summary.

## Validation

- `sh -n` + `go test ./tests/mw-dev` (the CI harness-script check) green.
- Setup/GC/teardown logic driven end-to-end against faked psql/kubectl: late-appearing password poll, stale DBs (>24h) dropped, fresh / unparseable / protected names untouched, run DB created, teardown drop verified gone.
- The change is self-exercising in CI: every e2e run creates, uses (full DuckLake activation + the reshard ext→cnpg copy now runs against a fresh, small catalog), and drops its database.

### DIFF
--- a/controlplane/provisioner/reshard_runner.go
+++ b/controlplane/provisioner/reshard_runner.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/posthog/duckgres/controlplane/configstore"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
@@ -61,6 +62,13 @@ type ReshardRunner struct {
 	// loopPoll is the inner-loop poll cadence (drain checks, flip waits,
 	// cancel checks). 5s in production; tests shrink it.
 	loopPoll time.Duration
+	// progressLogInterval is how often a wait loop (drain, flip converge,
+	// orphan-policy wait, recovery) writes what it currently observes to the
+	// op log. 15s in production; tests shrink it. Every wait loop MUST emit
+	// periodic observations at this cadence — a silent multi-minute poll loop
+	// is undebuggable from the op log (the mw-dev recovery that spun for ~8
+	// minutes on SASL auth failures with zero diagnostics).
+	progressLogInterval time.Duration
 
 	// extPasswords holds the EPHEMERAL external passwords by op id: the
 	// cnpg→ext target password (pulled from the creating CP replica's stash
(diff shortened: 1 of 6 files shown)


benben:
Base retargeted to `main`: the e2e workflow only triggers on PRs targeting main, and the whole point of this PR is to prove itself (and the stacked #950 commits it contains) against the real cluster. The diff currently shows #950's commits too; once #950 merges, this shrinks to the harness-side per-run-DB change alone.

PR #982, opened 2026-07-22
feat(auth): scoped read-only secret for external writers

jghoman:
## Problem

The discovery endpoints (`GET /api/v1/warehouses`, `GET /api/v1/warehouse-team-ids`, #966) are consumed by external writer pods — millpond's include-values poller today, viaduck's destination discovery next. They currently authenticate with the internal secret, which resolves to the **admin role for the entire API surface**: provisioning CRUD, password resets, impersonation, billing. A compromised writer pod holds a full CP-admin credential to read a team-id list. (Flagged in review on the millpond wiring PR.)

## Change

A scoped **read-only secret** whose blast radius is exactly the two discovery GETs (named for its scope, not the endpoint family — future read-only surfaces can join the same group):

- `--read-only-secret` / `DUCKGRES_READ_ONLY_SECRET` (+ `-fallbacks` rotation pair), mirroring internal-secret resolution semantics exactly
- Discovery routes move into their **own gin group** (`registerReadOnlyGroup` — the single mount point) behind `admin.AnyTokenAuthMiddleware`: header-only (no cookie fallback), no SSO, no role resolution; accepts the read-only token **or** the admin token (operator/debug + rotation window). The admin surface never accepts the read-only token.
- **Unset secret = exact pre-rollout behavior** (empty TokenSet never validates; admin token keeps working on discovery) — deploy ordering vs. the chart wiring is free, mixed fleets are safe.
- **Startup fails if the read-only value collides with the internal secret or any fallback** (`validateDistinctReadOnlySecret`) — a shared value would silently un-scope the credential.

## Review

Two adversarial review rounds (lead-QE + principal-SWE), both SHIP; both MEDs landed in-PR:
- QE: no tripwire pinned the real topology → `registerReadOnlyGroup` extraction + `TestReadOnlyGroupTopology` (exact route set + auth matrix on the real wiring; a new route on the group fails the test)
- SWE: same-secret collision silently defeats the scoping → the startup guard
- Plus: header-only hardening (cookie-negative test), truthful unset-secret logging keyed off `TokenSet.Count()` with `fallback_count`, config-resolution tests cloned from the internal-secret suite, CLAUDE.md documents the honest blast radius (tenant list + connection topology: RDS endpoints, bucket names, Secret *names* — never values)

Verified clean in review: no escalation path (gin group middleware binds per-route at registration; login can't mint a cookie from the read-only token), constant-time token compare unchanged, no Identity/audit dependency on the moved GETs (GETs were never audited; the admin SPA never calls these endpoints).

## Rollout state

1. ✅ ASM: `read-only-secret` + `read-only-secret-fallbacks` properties live in all three env tokens secrets (values hash-verified preserved)
2. posthog-cloud-infra#9537 — birth schema (`token_names`) for rebuilds/new envs
3. PostHog/charts#13380 — tokens ExternalSecret + `DUCKGRES_READ_ONLY_SECRET` env on the CP
4. Millpond ES property flip `internal-secret` → `read-only-secret` after the prod CP runs this + the charts wiring. Rollback note for after that flip: a CP rolled back to pre-this-PR code 401s discovery polls — consumers coast on last-known-good; new-tenant pickup stalls until re-roll.

No ordering constraints among 2/3 and this PR.

## Testing

`go build -tags kubernetes ./...`; full `go test -tags kubernetes ./controlplane/... ./configresolve/ ./cmd/...` green. New: topology tripwire, 8-case scoping matrix, cookie-negative, unset-secret, 7-case collision guard, 5 resolution tests.


benben (approved):

controlplane/multitenant.go
```diff
@@ -507,6 +507,24 @@ func SetupMultiTenant(
...
+	// that's a startup failure, not a warning.
+	if err := validateDistinctReadOnlySecret(cfg.ReadOnlySecret, cfg.ReadOnlySecretFallbacks, internalSecret, cfg.InternalSecretFallbacks); err != nil {
+		return nil, nil, nil, nil, nil, nil, err
+	}
+	readOnlyTokens := admin.NewTokenSet(cfg.ReadOnlySecret, cfg.ReadOnlySecretFallbacks)
+	if readOnlyTokens.Count() == 0 {
+		// Keyed off the TokenSet, not just cfg.ReadOnlySecret: fallbacks
+		// alone (mid-rotation) still validate, and saying otherwise here
+		// would mislead an operator debugging exactly that state.
+		slog.Info("Discovery secret not set; discovery endpoints accept only the internal secret. Set --read-only-secret or DUCKGRES_READ_ONLY_SECRET to give external writers a scoped credential.")
```
benben:
```suggestion
		slog.Info("Read-only secret not set; discovery endpoints accept only the internal secret. Set --read-only-secret or DUCKGRES_READ_ONLY_SECRET to give external writers a scoped credential.")
```

PR #986, opened 2026-07-22
ci: make ECR sha push + manifest idempotent

Piccirello:
## Summary

Make the ECR push and multi-arch manifest steps idempotent so re-running / `workflow_dispatch`-ing this workflow for an already-built commit doesn't fail — a prerequisite for turning on ECR tag immutability for `duckgres` (PostHog/posthog-cloud-infra#9549).

Today the `build` job pushes `<sha>-amd64` / `<sha>-arm64` and the `manifest` job creates `<sha>`. Once those tags are immutable, a re-run re-pushing them fails with `ImageTagAlreadyExistsException` in the `build` job, which skips the `manifest` job — so the **deploy dispatch to charts never fires**.

## Changes

- **build job:** a preflight `batch-get-image` check skips `Build and push by digest` when `<sha>-<arch>` already exists in ECR.
- **manifest job:** `Create and push ECR manifest` skips `imagetools create` when `<sha>` already exists, and resolves the digest from the existing (or freshly created) manifest either way — so the deploy dispatch still runs on a re-run.

Uses `batch-get-image` (the publish role has `ecr:BatchGetImage`, not `ecr:DescribeImages`). Normal first builds are unchanged.

### DIFF
--- a/.github/workflows/container-image-cd.yml
+++ b/.github/workflows/container-image-cd.yml
@@ -59,8 +59,33 @@ jobs:
               env:
                   PLATFORM: ${{ matrix.platform }}
 
+            # Skip the build+push when this commit's arch image is already in ECR, so a
+            # re-run/dispatch doesn't collide with the now-immutable <sha>-<arch> tag.
+            - name: Check if image already in ECR
+              id: check
+              env:
+                  SHA: ${{ github.sha }}
+                  ARCH: ${{ steps.slug.outputs.arch }}
+              run: |
+                  set -euo pipefail
+                  if aws ecr batch-get-image \
+                      --repository-name "$IMAGE_NAME" \
+                      --image-ids imageTag="${SHA}-${ARCH}" \
+                      --accepted-media-types \
+                          "application/vnd.oci.image.index.v1+json" \
+                          "application/vnd.oci.image.manifest.v1+json" \
+                          "application/vnd.docker.distribution.manifest.list.v2+json" \
+                          "application/vnd.docker.distribution.manifest.v2+json" \
+                      --query 'images[0].imageId.imageDigest' --output text 2>/dev/null | grep -q '^sha256:'; then
+                      echo "exists=true" >> "$GITHUB_OUTPUT"
+                      echo "Image ${SHA}-${ARCH} already in ECR; skipping build."
+                  else
+                      echo "exists=false" >> "$GITHUB_OUTPUT"
+                  fi
+
             - name: Build and push by digest
               id: build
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image-cd.yml
```diff
@@ -59,8 +59,33 @@ jobs:
               env:
                   PLATFORM: ${{ matrix.platform }}
 
+            # Skip the build+push when this commit's arch image is already in ECR, so a
+            # re-run/dispatch doesn't collide with the now-immutable <sha>-<arch> tag.
+            - name: Check if image already in ECR
```
Z3r0Sum, line 64: I wonder if some of this stuff is worth codifying in a more generic way here: https://github.com/PostHog/shared-actions/blob/main/container-image-metadata/action.yml ... and potentially replacing some of the dupe stuff below.  Outside the scope of the PR probably, but kinda makes sense imho.

Z3r0Sum (approved):

.github/workflows/container-image-cd.yml
Piccirello: Agreed — as more repos flip to immutable ECR, every build workflow ends up needing this same skip-if-exists guard, and this workflow already repeats the buildx / AWS-creds / ECR / GHCR-login boilerplate across both jobs, so a shared composite action makes sense.

One nuance: `container-image-metadata` is purpose-built to emit the tags/labels/OCI-annotation provenance contract (no registry I/O), so the existence-check + push/manifest logic fits better as a *sibling* action that composes with it than baked inside it.

Keeping it out of this PR though — this is a minimal unblock so re-runs survive immutability, and the extraction spans shared-actions plus every consumer. Good case for it given it's now hand-written here (build + manifest) and in posthog's `container-images-cd`.

Z3r0Sum: > Agreed — as more repos flip to immutable ECR, every build workflow ends up needing this same skip-if-exists guard, and this workflow already repeats the buildx / AWS-creds / ECR / GHCR-login boilerplate across both jobs, so a shared composite action makes sense.
> 
> One nuance: `container-image-metadata` is purpose-built to emit the tags/labels/OCI-annotation provenance contract (no registry I/O), so the existence-check + push/manifest logic fits better as a _sibling_ action that composes with it than baked inside it.
> 
> Keeping it out of this PR though — this is a minimal unblock so re-runs survive immutability, and the extraction spans shared-actions plus every consumer. Good case for it given it's now hand-written here (build + manifest) and in posthog's `container-images-cd`.

Yes robot Tom, that was just an example 🤣

PR #1004, opened 2026-07-28
feat(analytics): emit per-org query_completed event

annaszell:
## What

Adds a `query_completed` product-analytics event, emitted per org from `logQuery` on the terminal event of a **successful** statement. Props: `user`, `trace_id`, `protocol`, `query_kind`, `duration_ms`, `cpu_seconds` (DuckDB CPU/thread-time from the profiling rollup), `result_rows`.

It sits alongside the existing `query_initiated` / `query_failed` events and is gated on `errCode == ""`, so failures stay covered by `query_failed` and the lifecycle reads `initiated -> completed | failed` with no double-counting.

## Why

PostHog needs a per-org "the warehouse was actually used" signal to define managed-warehouse **activation** (e.g. queried on ≥2 distinct days within 14 days of provisioning, or cumulative `cpu_seconds` over a threshold). `query_initiated` gives us query starts, but not success or resource cost. The old `managed warehouse compute usage` push event that carried `cpu_seconds` was retired in #893 when billing moved to the pull API, so that signal no longer reaches PostHog. This adds it back as a lightweight product-analytics event (not a billing path), consistent with the `query_initiated`/`query_failed` events the pull-API migration kept.

## Implementation notes

- The profiling rollup (`lastProfilingSummary`) is now consumed at the **top** of `logQuery`, so the analytics event and the query-log row observe the same values, and it is reset even when the query-log sink is disabled.
- The event is emitted **independently of the query-log sink**, so this usage signal does not depend on query-log configuration.
- Capture is async and batched (existing `analytics` tracker), so it stays off the query latency path.

> [!NOTE]
> **Granularity, please sanity-check.** `query_completed` fires at *statement* granularity (it lives in `logQuery`, which covers simple, extended, simple-batch, COPY, and cursor statements). So a single logical client request can emit more than one `query_completed` — e.g. cursor FETCHes or COPY batches — unlike `query_initiated`, which fires once per accepted request. I chose statement granularity so each statement's `cpu_seconds`/`duration_ms` is captured; downstream filters by `query_kind` to isolate real data queries from utility statements. If you'd rather have a 1:1 pairing with `query_initiated`, say so and I'll rework it.

> [!NOTE]
> **Scope: pgwire only.** The Flight SQL ingress is disabled by default (`flight-port 0`) and has no per-query resource plumbing today, so it's out of scope here. Wiring an equivalent event on that path is a follow-up if/when it's enabled for managed warehouse.

## Testing

- `go build ./server/`, `go vet ./server/`, `golangci-lint run ./server/` — all clean.
- Added `TestLogQueryEmitsQueryCompletedOnSuccess` and `TestLogQueryDoesNotEmitCompletedOnError` to `conn_analytics_test.go`; existing `logQuery`/analytics tests still pass (the profiling-rollup restructure is covered by `TestLogQueryCopiesProfilingSummaryToQueryLogEntry`).
- I (Claude, via PostHog Code) could not run the warehouse end-to-end; verification is the unit tests above plus build/vet/lint.

Updated the events table + notes in `README.md`.

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,14 @@ the `distinct_id` is the org name and each event carries a group of type
 single-tenant standalone mode (no org) the `distinct_id` is `standalone` and no
 group is attached.
 
+The org name is duckgres-internal, so the query events additionally carry a
+`team_id` property — the PostHog `Team.id` for the connection (the connecting
+user's team, else the org's oldest team; 0 when unknown or standalone). This is
+the PostHog-native key that joins duckgres usage to the rest of PostHog (e.g.
+product-intent cohorts for managed-warehouse activation). It is a config-snapshot
+read stamped once per connection, and mirrors the informational team id the
+compute-usage meter records.
+
 Events never include SQL text, credentials, or secret values — only metadata.
 
 Provisioning and deprovisioning are asynchronous: the admin API returns `202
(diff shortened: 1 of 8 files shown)


annaszell:
Heads-up on a join-key gap I found after opening this (flagging rather than silently shipping):

To use `query_completed` for the PostHog managed-warehouse **activation** metric, we need to join it to `user showed product intent` cohorts. But those are keyed by PostHog **org UUID / team_id**, whereas duckgres analytics attribute by the internal **org slug** (`distinct_id = orgID`, a DNS-1123 slug ≤35 chars). The same mismatch affects the `organization` group these events set (keyed by slug, not the PostHog org id). So as-is, this event can't be joined to the rest of PostHog.

Options, roughly:
1. Stamp PostHog `team_id` onto the connection at setup and include it on `query_completed` (and ideally `query_initiated`). duckgres already resolves this for billing via `resolveTeam` / the org's `default_team_id`, so the value is available — it's a matter of plumbing it onto `clientConn`. This feels like the right fix and would make all duckgres product analytics joinable, but it touches connection setup, so I'd want your steer before extending this PR.
2. Bridge on the PostHog side by persisting the duckgres org slug at provision time and joining on it.

Happy to extend this PR with (1) if you're on board with adding `team_id` to the analytics events — or if you'd prefer a different identifier, let me know what's cleanly available.

benben (approved):

annaszell:
Resolved the join-key gap in 22b46a7: the query events now carry a `team_id` property (PostHog `Team.id`), stamped on the connection at control-plane setup via the same `OrgUsageTeamID` resolver the compute meter uses (config-snapshot read, no I/O — connecting user's team, else the org's oldest team, else 0). Added `team_id` to `query_initiated`, `query_completed`, and `query_failed` so all three join cleanly to PostHog org/team, plus `OrgUsageTeamID` on `ConfigStoreInterface` and the test fakes. README updated.

Still happy to hear if you'd prefer a different identifier or placement.

PR #1012, opened 2026-07-30
feat(duckdbservice): export ducklake commit-loop stats from ducklake_commit_stats()

fuziontech:
## Summary

Workers now poll the ducklake extension's (upcoming) `ducklake_commit_stats()` table function every 30s and re-export its process-global counters as Prometheus metrics. This gives visibility INSIDE the extension's commit retry loop — the existing `duckgres_worker_ducklake_conflict_*` counters only see conflicts that escape it, which is why they read ~0 while megaduck's metadata store shows a 12.4% transaction rollback rate.

New metrics (all labelled by `catalog`; conflicts additionally by `cause` ∈ {primary_key, unique, conflict, concurrent}):

- `duckgres_worker_ducklake_commit_attempts_total`
- `duckgres_worker_ducklake_commit_successes_total`
- `duckgres_worker_ducklake_commit_retries_exhausted_total`
- `duckgres_worker_ducklake_commit_nonretryable_errors_total`
- `duckgres_worker_ducklake_commit_backoff_ms_total`
- `duckgres_worker_ducklake_commit_duration_ms_total`
- `duckgres_worker_ducklake_commit_conflicts_total`

## Design

- **Contract** (shared with the extension-side PR in PostHog/ducklake): `ducklake_commit_stats()` → rows of `(catalog VARCHAR, stat VARCHAR, value BIGINT)`, cumulative and monotonic per process. Cumulative→counter conversion via last-seen delta tracking (decrease → treated as fresh baseline; negative values → dropped, never stored as baseline).
- **Graceful degradation**: capability probe (`duckdb_functions()`) once per engine; no released extension binary has the function yet, so on current pins this logs once at info and stays dormant. Ships safely before/independently of the extension PR.
- **Lifecycle**: `commitStatsLoop` mirrors `metadataMetricsLoop` — started in `NewDuckDBService`, 30s ticker, stops via the pool's `stopCh`, scrapes over `controlDB` to avoid queueing behind pinned user-query connections.

## Test plan

- [x] 6 new unit tests: delta table (incl. decrease/negative), stat→metric mapping (incl. unknown-cause passthrough + unknown-stat rejection), end-to-end record via `prometheus/testutil`, probe against real in-memory DuckDB (macro stand-in), probe-once/disabled path, nil-DB skip
- [x] `just test-unit` (incl. repository hygiene), `just lint`, `go vet`: clean
- [ ] e2e positive path: not exercisable until an extension binary ships `ducklake_commit_stats()` (noted per harness rule); degradation path is what current pins exercise

Follow-ups: millpond needs the same poller (separate repo — that's the writer where megaduck's numbers will appear); Grafana panels once metrics flow.

https://claude.ai/code/session_01LUnQHp46rmLKdupmDUr7kG

### DIFF
--- a/duckdbservice/commit_stats.go
+++ b/duckdbservice/commit_stats.go
@@ -1,0 +1,257 @@
+package duckdbservice
+
+// Polls the posthog ducklake extension's ducklake_commit_stats() table
+// function and re-exports its counters as the
+// duckgres_worker_ducklake_commit_* Prometheus metrics (metrics.go).
+//
+// The function contract (shared with the extension):
+//   - Signature: ducklake_commit_stats() — no arguments.
+//   - Columns: catalog VARCHAR, stat VARCHAR, value BIGINT.
+//   - One row per (catalog, stat); values are process-global, cumulative, and
+//     monotonic for the life of the process.
+//   - Stat names: attempts, successes, retries_exhausted, nonretryable_errors,
+//     backoff_ms, total_commit_ms, and conflicts.<cause> with cause one of
+//     primary_key, unique, conflict, concurrent (an unknown cause from a newer
+//     extension is exported under its own cause label rather than dropped).
+//
+// The function only exists in ducklake extension builds that expose the
+// commit retry loop counters — no currently-released extension binary does —
+// so availability is probed once per DuckDB engine and, when absent, polling
+// is disabled with a single info log. Everything degrades gracefully with the
+// pinned extension.
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"log/slog"
+	"strings"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	commitStatsInterval      = 30 * time.Second
+	commitStatsScrapeTimeout = 5 * time.Second
+
+	commitStatsProbeQuery = "SELECT 1 FROM duckdb_functions() WHERE function_name = 'ducklake_commit_stats' LIMIT 1"
+	commitStatsQuery      = "SELECT catalog, stat, value FROM ducklake_commit_stats()"
+
+	commitStatsConflictPrefix = "conflicts."
+)
+
+// commitStatRow is one (catalog, stat, value) row from ducklake_commit_stats().
+type commitStatRow struct {
+	Catalog string
+	Stat    string
+	Value   int64
+}
+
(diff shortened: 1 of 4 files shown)


fuziontech (commented):
Adversarial self-review (requested by James). One real bug found and fixed in be2e015; the rest verified clean.

**Bug found & fixed:** the delta tracker's defensive path for negative values reset the baseline to 0 (`lastSeen[key] = 0`), which would make the *next valid* cumulative value count in full — double-exporting everything already recorded (100 exported, glitch, 102 arrives → +102 instead of +2). The inline comment claimed this prevented inflated deltas; it caused them. Fix: keep the baseline untouched and emit nothing. The test now regression-covers the exact double-count scenario ("post-glitch value counts only the increase over the kept baseline"). Contract-wise negatives can't occur (extension counters are monotonic int64 atomics), so this was latent, but a defense that amplifies is worse than none.

**Verified correct:**
- Probe caching is by `*sql.DB` identity and a probe *error* records no verdict (retries next tick) — no risk of permanently wedging into "unavailable" from a transient failure; engine replacement triggers exactly one re-probe.
- Exactly one poller goroutine per process: `NewDuckDBService` is invoked once in the service path, and the exporter state is confined to that goroutine (no locking needed, as documented).
- `scrapeCommitStats` takes `p.mu.RLock` only to read the `controlDB` pointer, never across a query; 5s scrape timeout; falls back to activation DB only when `controlDB` is unwired (test paths).
- Counter-reset semantics: decrease → current value as delta is the standard restart heuristic; correct here since exporter state and extension counters share process lifetime.
- Zero-delta touches materialize all known series on first poll (good for `rate()` continuity).

**Notes for reviewers (accepted trade-offs):**
1. `*_ms_total` deviates from the Prometheus base-unit (seconds) convention — kept deliberately for 1:1 symmetry with the extension contract and integer delta math. `rate()` dashboards just divide by 1e3.
2. Unknown `conflicts.<cause>` values export under their own label (forward-compatible with new extension causes; cardinality is extension-controlled — ours). Unknown non-conflict stats are dropped.
3. The positive e2e path can't run until an extension binary ships `ducklake_commit_stats()` (PostHog/ducklake#35 → next release tag); until then prod exercises exactly the probe-absent path, which is unit-tested against real DuckDB.

Post-fix: `go build`, `go vet`, unit tests, `just lint` all clean.

PR #1020, opened 2026-07-30
build(docker): switch runtime images from debian to wolfi-base

feliperalmeida:
Switches the runtime stage of all three images (`Dockerfile`, `Dockerfile.controlplane`, `Dockerfile.worker`) from `debian:bookworm-slim` to `chainguard/wolfi-base`. Builder stages are unchanged.

- **PG-18 client without the PGDG bootstrap**: Wolfi carries `postgresql-18-client` natively, so the wget/gnupg/apt-repo dance for the reshard runner's `pg_dump`/`pg_restore` is gone (pin to PG 18 preserved).
- **`libstdc++` added explicitly** where the CGO-linked DuckDB engine needs it (`Dockerfile`, `Dockerfile.worker`) — debian-slim shipped it implicitly, wolfi-base doesn't. Deliberately omitted from the CP image, which doesn't link libduckdb.
- Non-root `duckgres` user now created via busybox `addgroup`/`adduser`.
- Image sizes: all-in-one 660→535MB, CP 292MB, worker 497MB.

Verified locally: all-in-one serves psql over TLS with DDL/DML + JSON-extension queries; worker pre-warms DuckDB and loads the bundled ducklake extension; CP binary runs with `pg_dump` 18.4 present; all run as non-root on Wolfi.

### DIFF
--- a/Dockerfile.controlplane
+++ b/Dockerfile.controlplane
@@ -43,20 +43,14 @@ RUN CGO_ENABLED=1 go build -tags "${BUILD_TAGS}" \
     -o duckgres-controlplane \
     ./cmd/duckgres-controlplane
 
-FROM debian:bookworm-slim
+FROM chainguard/wolfi-base:latest
 
-# postgresql-client-18 provides pg_dump/pg_restore for the reshard runner's
+# postgresql-18-client provides pg_dump/pg_restore for the reshard runner's
 # pre-flip catalog backup (docs/design/resharding.md). Pinned to PG 18 to match
-# the cnpg shard major so it can dump PG-18 catalogs. From the PGDG apt repo
-# because Debian bookworm's stock repo only carries PG 15.
-RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates gnupg wget \
-    && install -d /usr/share/postgresql-common/pgdg \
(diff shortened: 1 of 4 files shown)


Dockerfile.controlplane
```diff
@@ -43,20 +43,14 @@ RUN CGO_ENABLED=1 go build -tags "${BUILD_TAGS}" \
     -o duckgres-controlplane \
     ./cmd/duckgres-controlplane
 
-FROM debian:bookworm-slim
+FROM chainguard/wolfi-base:latest
```
Piccirello, line 46: Is there any risk in moving from a specific version of debian to unpinned wolfi version?

Piccirello (approved):
Security side LGTM but defer to MW for signoff

Dockerfile.controlplane
feliperalmeida: I thought about that for a bit, and ended up concluding that [wolfi-base](https://edu.chainguard.dev/open-source/wolfi/overview/) is such a small image (basically OS + glibc + apk + bash) that seemed ok to use `latest` - they provide [sha tags](https://hub.docker.com/r/chainguard/wolfi-base/tags) for reproducible builds if that's something we'd look for, but I don't think that should be needed.

bill-ph (approved):

PR #1026, opened 2026-08-03
feat(controlplane): serve persons_distinct_ids_table in discovery payload

fuziontech:
## What

Adds `persons_distinct_ids_table` to the per-team payload served by `GET /api/v1/warehouses` (destination discovery for external writers).

## Why

Streamed persons replication (millpond → shared DuckLake changelog → viaduck `full_cdc` fanout) lands **two** tables per team in the duckling: the persons table and its distinct-id mapping table (the join side of today's denormalized batch export). Discovery already resolves `events_table`/`persons_table`; the distinct-ids location belongs in the same single derivation site so a viaduck persons pipeline never re-derives the suffix rule itself.

## Derivation rule

`resolveTeamTables` now also returns `<schema_name>.` + the resolved persons name with its `persons` prefix replaced by `persons_distinct_ids`:

- `persons` → `persons_distinct_ids`
- `persons_ab12` (suffixed override) → `persons_distinct_ids_ab12`
- an override without the `persons` prefix → `<name>_distinct_ids`

No new configstore column: this table has no legacy/grandfathered shape, so convention-only derivation is sufficient.

## Compatibility

Additive field on the existing team payload — wire-compatible per the discovery module's own contract (adding a field is safe, removing one is not).

## Tests

- `TestResolvedTableNames` extended to assert the default derivation
- New `TestResolvedDistinctIDsTableFollowsPersonsSuffix` for the suffixed-override case
- `go test ./controlplane/...` green; `golangci-lint run` clean

🤖 Prepared by an agent as part of the persons-backfill → millpond/viaduck streaming migration.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -762,6 +762,10 @@ are load-bearing for those consumers:
   reads as fleet-wide removal to every consumer).
 - **Teams come from `duckgres_org_teams`**, with RESOLVED table locations:
   `events_table`/`persons_table` = `<schema_name>.<override-or-derived>`,
+  `persons_distinct_ids_table` = `<schema_name>.` + the resolved persons
+  name with its `persons` prefix replaced by `persons_distinct_ids` (so a
+  suffixed override stays disambiguated: persons_ab12 →
+  persons_distinct_ids_ab12),
   `data_imports_schema` = override-or-`<schema>_data_imports`. The
   derivation lives ONCE, in `resolveTeamTables`; the legacy overrides are
   BARE identifiers (never schema-qualified), enforced at every write surface
--- a/controlplane/provisioning/discovery.go
+++ b/controlplane/provisioning/discovery.go
@@ -4,6 +4,7 @@ import (
 	"log/slog"
 	"net/http"
 	"sort"
+	"strings"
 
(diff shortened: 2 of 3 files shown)


fuziontech:
Downstream consumer of this field: PostHog/posthog#76997 (duckling persons streaming schema prep + cutover runbook) derives the same name with `persons_distinct_ids_table_name()`, documented as rule-for-rule identical to `distinctIDsTableName` here. A prepared viaduck branch adds `discovery.table_field` so a persons pipeline can select `persons_table` / `persons_distinct_ids_table` from this payload.

fuziontech:
Full PR map for the persons streaming migration:

- PostHog/duckgres#1026 (this PR) — discovery payload field
- PostHog/viaduck#67 — `discovery.table_field` consuming it
- PostHog/posthog#76997 — duckling schema prep + cutover runbook
- PostHog/charts#13846 — viaduck persons pipeline ApplicationSets (shipped disabled; static-destination cutover, discovery flip is post-fleet-migration)

Note: the millpond side needs nothing — its `person` / `person-distinct-id` consumers already land the changelogs on megaduck in prod.

PR #1038, opened 2026-08-05
fix(protocol): portal suspension for extended-query Execute row limits (Hex 1024-row truncation)

EDsCODE:
## Problem

A customer reported that Hex caps every result set at exactly **1,024 rows** — `select * from ducklake.posthog.persons_production limit 3000` returned 1,024 rows with no error. Hex confirmed they apply no client-side limit.

Root cause: duckgres did not implement **portal suspension**. Hex's driver (like JDBC with `setFetchSize`) pages result sets by sending `Execute` with a row limit of 1024. Per the PostgreSQL protocol, a server that hits the limit with rows remaining must send `PortalSuspended` so the client issues another `Execute` for the next page. Duckgres instead sent `CommandComplete` after the first page (`server/conn_extended_query.go`: `// Portal suspended - but we don't support this yet`), so the client — correctly, per protocol — concluded the result set was complete. Silent truncation at exactly the client's page size, for any paging client.

Reproduced against a local standalone build with a raw `pgproto3` client: `Execute{MaxRows: 1024}` on a 3,000-row query → 1,024 `DataRow`s + `CommandComplete "SELECT 1024"`.

## Fix

Implement portal suspension per the protocol:

- Hitting the row limit sends **`PortalSuspended`** and keeps the portal's rowset open (`portal.exec`); a follow-up `Execute` on the same portal **resumes** streaming where the previous leg stopped — the query executes exactly once across legs. The limit is now checked *before* advancing the rowset, fixing a latent off-by-one where the first row of the next page was consumed and dropped.
- `CommandComplete` is sent only when the rowset is exhausted, with the cumulative row count (PostgreSQL semantics: a limit equal to the result size still suspends; the next `Execute` returns 0 rows + complete).
- **Cleanup**: suspended portals are destroyed (rowset closed, portal deleted) at transaction end — mirroring `closeCursorsAtTxEnd`, both for PostgreSQL parity and because the open rowset pins the session's single DuckDB connection. `Close('P')`, Bind-overwrite of the portal name, and connection teardown also release the rowset. Destroying rather than keeping the tx-ended portal matters: a kept entry would re-run the query from row 0 on the next `Execute`, silently replaying page one as a continuation; `34000 "portal does not exist"` is the honest answer.
- The query-log entry is written by the leg that completes (or errors) the portal, spanning all legs with the cumulative row count.

## Verification

- **Unit** (`server/conn_portal_suspension_test.go`, written first, watched fail): exact wire-message sequences for suspend/resume paging (`TDDs`/`DDs`/`DC` + `SELECT 5` tag, query executed once), exact-boundary suspension, `maxRows=0` regression guard, and rows-closed asserts for Close/tx-end/Bind-overwrite.
- **Integration** (`tests/integration/portal_suspension_test.go`): raw `pgproto3` frontend pages a 1,000-row result set in 128-row pages against the real server over TLS — passes with all 1,000 rows, ≥7 suspensions, cumulative `SELECT 1000` tag.
- **End-to-end repro**: the original failing scenario (3,000 rows, `MaxRows: 1024`) against a locally running standalone build now returns `PortalSuspended` at 1,024 and 2,048 and all 3,000 rows with `CommandComplete "SELECT 3000"`.
- `go build ./...`, `go test ./server/... ./transpiler/... ./duckdbservice/... ./controlplane/` all pass; cursor/JDBC/session integration subsets pass.
- **mw-dev harness**: not assertable in-Job — the harness drives SQL through psql, and libpq never sends a nonzero Execute row limit. Noted under "Deliberately not covered here" in `tests/mw-dev/README.md` per the testing policy; the suspension path lives entirely in the CP's pgwire layer above the executor, so it is identical on the remote backend.

Docs: `docs/postgres-compatibility.md` wire-protocol table gains a Portal suspension row.

### DIFF
--- a/docs/postgres-compatibility.md
+++ b/docs/postgres-compatibility.md
@@ -163,6 +163,7 @@ Operational guidance: allow the import to complete, then manually deduplicate af
 |---------|--------|---------|-------|
 | Simple query | ✅ | `protocol_test.go::TestProtocolSimpleQuery` | |
 | Extended query (Parse/Bind/Describe/Execute/Sync) | ✅ | `protocol_test.go::TestProtocolExtendedQuery`; server `conn_bind_test.go`, `conn_describe_test.go` | |
+| Portal suspension (Execute row limit / PortalSuspended) | ✅ | `portal_suspension_test.go::TestPortalSuspensionPaging` (raw-frontend paging); server `conn_portal_suspension_test.go` | Paging clients (JDBC `setFetchSize`, Hex) resume with repeat Execute; the query runs once. Suspended portals are destroyed at transaction end (PostgreSQL parity + the same single-connection liveness rule as cursors). Pre-fix, the first page was answered with CommandComplete — silent truncation at the client's page size |
 | Extended-query error recovery (skip-until-Sync, pipelining) | ✅ | server `conn_skip_until_sync_test.go`; clients `clients_test.go::TestExtendedQueryErrorHandling` | #718 |
 | Prepared statements (PREPARE/EXECUTE, reuse, NULL, 20+ params) | ✅ | `edge_cases_test.go::TestPreparedStatementEdgeCases`; clients `::TestPreparedStatements` | |
 | Binary vs text result format | ✅ | `protocol_test.go::TestProtocolDataTypes`; clients `::TestPgxBinaryFormatResults`; server `types_test.go` encode/decode | |
--- a/server/conn.go
+++ b/server/conn.go
@@ -110,6 +110,36 @@ type portal struct {
 	paramFormats  []int16 // 0=text, 1=binary for each parameter
 	resultFormats []int16
 	described     bool // true if Describe was called on this portal
+	// exec is set while the portal is suspended: Execute hit its row limit
+	// (PortalSuspended sent) with the result set unexhausted. The next
+	// Execute on this portal resumes streaming from exec.rows instead of
+	// re-running the query.
+	exec *portalExec
+}
+
+// portalExec is the streaming state of a suspended portal: the still-open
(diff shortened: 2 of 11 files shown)


EDsCODE:
Rebased onto latest main and resolved the conflict with #1034 (exploratory small-worker tier), which had refactored the Execute row loop into the shared `streamSelectRows` helper. Resolution notes:

- **`streamSelectRows` is now suspension-aware** (single shared streamer, keeping #1034's intent): new `limitReached` outcome, and the limit is checked *before* advancing the rowset — fixing the dropped-row behavior its own doc comment noted ("the row already fetched when the cap is reached is dropped"). The simple-query path passes `maxRows=0` and is unchanged.
- **The zero-row OOM retry composes with suspension**: the retried rowset (`activeRows`) is the one a suspension keeps open, and its deferred close respects `keepRowsOpen`. No overlap in triggers — the retry needs `rowsErr` with 0 rows sent; suspension implies a clean stream. Resume legs deliberately have no retry (rows from earlier legs are already out the door).
- **`escalateWorker` now destroys suspended portals** after a successful swap: the switcher destroys the old session, so their rowsets died with it — a later Execute on one gets an honest `34000` instead of a dangling rowset.
- #1034's `TestExtendedExecuteMaxRowsUnchanged` pinned the old truncate-at-cap behavior; it's now `TestExtendedExecuteMaxRowsSuspends` asserting suspend + resume + cumulative tag through the tier harness.
- Resume legs open their own query-metrics scope (per-Execute-message semantics), which also gives them the terminal wire flush every Execute leg relies on.

All server/transpiler/duckdbservice/controlplane suites plus the integration paging test pass on the rebased branch.

PR #1047, opened 2026-08-09
perf(cache-proxy): bound, hedge, and adapt peer fetches

EDsCODE:
## Problem

Peer lookup is useful only while it beats origin. The previous implementation had three failure modes under contention: peer fills were unbounded across requests, a fixed wait delayed origin even when the fleet was degraded, and losing transfers could continue for up to 30 seconds after the response path had already chosen the other source.

## Changes

- **Bound peer work process-wide.** Keep the existing 8-fill per-request fairness limit and add shared count and byte ceilings. `CACHE_PEER_FETCH_MAX_CONCURRENCY` defaults to 32; `CACHE_PEER_FETCH_MAX_BYTES` defaults to concurrency × block size. Permit queueing consumes the hedge head start, so overloaded work sheds to the coalesced origin path instead of starting late.
- **Use a true adaptive hedge.** Origin starts while peer I/O is still running after a head start equal to the rolling p50 of the last 64 successful peer block fetches, clamped to 25–150 ms. Origin work continues through the existing miss-run coalescing and shared span flight rather than becoming one request per block.
- **Cancel the losing side.** Each validated origin block commit cancels the matching peer transfer for every waiter on the shared span. A peer win releases and cancels the origin span once no other request still needs it. Partial losing-side bytes are recorded as duplicate work.
- **Add a latency circuit breaker.** The proxy compares peer latency with time through the first atomically committed origin block. Eight sustained observations above 1.5× origin open the breaker; open proxies go directly to origin and run one bounded, non-blocking recovery probe every 5 seconds. Three healthy samples close it again. Abrupt slowdowns and origin-before-first-byte stalls have bounded diagnostic paths.
- **Expand observability.** Add hedge/winner, cancellation, duplicate-byte, late-success, breaker-transition, shed, queue/fetch latency, EWMA, and process-wide in-flight count/byte metrics. `Served.` lines now include the client address without its ephemeral port. Existing phase-timing fields and `blocks_hedged` remain intact.
- **Keep the proven mechanics.** Peer `has` probes retain their 150 ms timeout, block fills remain parallel, and all peer/body operations now honor caller cancellation.

Runtime defaults and a peer-path recovery runbook are documented in `cmd/cache-proxy/README.md`; the new metric families are documented in `docs/metrics.md`.

## Default-cap sweep

An isolated 16/32/64 contention sweep selected 32. A cap of 64 removed synthetic origin fallback but did not show a stable latency improvement and doubled the maximum byte reservation. A cap of 32 materially reduced fallback versus 16 while preserving the safer resource ceiling.

## Testing

- `just test-cache-proxy`
- Focused new-path tests under `go test -race`, 3 repetitions
- `just lint`

Coverage includes global count/byte admission, queued-deadline shedding, adaptive head starts, shared-flight and per-block cancellation, partial duplicate-byte accounting, mixed peer/origin spans, abrupt breaker opening, bounded recovery before and after the first origin block, pure-peer reset behavior, and client attribution.

### DIFF
--- a/cmd/cache-proxy/block_serve.go
+++ b/cmd/cache-proxy/block_serve.go
@@ -28,6 +28,10 @@ var (
 		Name: "cache_proxy_block_reads_total",
 		Help: "Blocks resolved while assembling responses, by source",
 	}, []string{"source"}) // local, peer, s3
+	peerFillHedgedTotal = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "cache_proxy_peer_fill_hedged_total",
+		Help: "Block fills sent to origin because the peer wait budget expired (the peer fetch continues in background)",
+	})
 	// requestDurationSeconds is shared between the block-serve path (this
 	// file) and the forward-proxy path (proxy.go); buckets start at 1ms
 	// because a local cache hit can be sub-millisecond and top out around 8s
@@ -178,6 +182,42 @@ func (p *CacheProxy) fetchOriginSpan(r *http.Request, blockSize, firstIdx, lastI
 	return nil
 }
 
+// peerFillConcurrency bounds how many of one request's blocks are fetched
+// from peers at the same time. Requests typically span one or two blocks, so
+// this only matters for wide spans, where it keeps a single request from
+// monopolizing peer bandwidth.
+const peerFillConcurrency = 8
+
+// peerFill is the handle for one block's background peer fetch: done closes
+// when the fetch finishes, and ok reports whether a peer delivered the block.
+type peerFill struct {
+	done chan struct{}
+	ok   bool
+}
+
+// waitFill waits for a peer fill until deadline. filled reports whether the
+// block is now on local disk; hedged reports that the budget expired first —
+// the block should be fetched from origin while the fill keeps running in the
+// background (a late fill still populates the cache for future requests).
+func waitFill(f *peerFill, deadline time.Time) (filled, hedged bool) {
+	select {
+	case <-f.done:
+		return f.ok, false
+	default:
+	}
+	if wait := time.Until(deadline); wait > 0 {
+		timer := time.NewTimer(wait)
+		defer timer.Stop()
+		select {
+		case <-f.done:
+			return f.ok, false
+		case <-timer.C:
+		}
+	}
+	peerFillHedgedTotal.Inc()
+	return false, true
+}
+
 // serveBlockAligned serves a cacheable GET whose Range is an absolute
 // bytes=start-end pair from block-aligned cache entries: local disk, then
 // peers, then coalesced origin fetches for contiguous missing runs (chunked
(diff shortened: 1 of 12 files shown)


cmd/cache-proxy/block_serve.go
```diff
@@ -218,9 +258,45 @@ func (p *CacheProxy) serveBlockAligned(w http.ResponseWriter, r *http.Request, r
...
+		fills = make(map[int64]*peerFill, blockCount)
+		sem := make(chan struct{}, peerFillConcurrency)
+		for idx := firstIdx; idx <= lastIdx; idx++ {
+			key := BlockKey(urlStr, idx, p.blockSize)
+			if p.store.Has(key) {
+				continue
+			}
+			f := &peerFill{done: make(chan struct{})}
+			fills[idx] = f
+			go func() {
```
bill-ph: [P1] This launches one goroutine per missing block; the semaphore only parks excess goroutines and is scoped per request. Because `blockCount` is bounded only by cache capacity (not `maxSpanBlocks`), a wide accepted range can retain a very large number of goroutines. Queued jobs also acquire slots after the 400 ms deadline and still call `FetchFromPeers` for up to 30 seconds even after the origin hedge has populated those keys. Please use a bounded/global worker queue or acquire before spawning, cancel expired jobs, and recheck `store.Has(key)` after acquiring. A >8-block test should assert no peer GETs start after the deadline.

cmd/cache-proxy/block_serve.go
```diff
@@ -277,32 +353,39 @@ func (p *CacheProxy) serveBlockAligned(w http.ResponseWriter, r *http.Request, r
...
-		if p.peers != nil {
+		// The fill check must precede the Has check: a block our own prefetch
+		// has already landed would otherwise pass Has and be counted as a
+		// local hit, corrupting the local/peer split in the log and metrics.
+		if f := fills[idx]; f != nil {
 			peerStart := time.Now()
-			_, _, ok := p.peers.FetchFromPeers(key, func(rd io.Reader) (int64, error) {
-				return p.store.PutStream(key, rd)
-			})
+			filled, hedged := waitFill(f, fillDeadline)
```
bill-ph: [P2] After this times out, the peer fill keeps running but its completion is no longer considered by the request. If the peer lands the block at 450 ms and the origin hedge fails at 500 ms, `flushRun` returns the origin error even though the requested block is valid locally (pre-PR, the peer would have served it). Before propagating an origin error, recheck the run's block keys/fill handles so either successful side of the hedge can satisfy the request.

cmd/cache-proxy/block_serve.go
```diff
@@ -178,6 +182,42 @@ func (p *CacheProxy) fetchOriginSpan(r *http.Request, blockSize, firstIdx, lastI
...
+	if wait := time.Until(deadline); wait > 0 {
+		timer := time.NewTimer(wait)
+		defer timer.Stop()
+		select {
+		case <-f.done:
+			return f.ok, false
+		case <-timer.C:
+		}
+	}
+	peerFillHedgedTotal.Inc()
```
bill-ph: [P2] This increments `cache_proxy_peer_fill_hedged_total` as soon as the deadline wins, but the caller subsequently checks `store.Has` and may avoid origin entirely if the peer or another request landed the block. That makes this counter and `blocks_hedged` overstate the documented “blocks fetched from origin” signal. Increment when the block is actually added to an origin miss run, or rename/document the metric as peer-budget expiration.

bill-ph (commented):
Reviewed the peer-path changes, including the cache-proxy suite and targeted race tests. The tests pass, but I found one high-priority background-work/resource issue and two hedge correctness/observability issues that are not covered by the new tests.

PR #1054, opened 2026-08-11
Allow mid-session SET of duckgres.worker_ttl

fuziontech:
## Why

On the remote/K8s backend, a worker whose last session ends is parked `hot_idle` (warm, quickly reusable) and retired once its TTL expires — 1 minute by default. The only way to extend that was the `-c duckgres.worker_ttl=...` connect-time startup option, which many clients (BI tools, poolers) can't set. Their workers are reaped at the default TTL whenever a connection goes idle, so every return visit pays a cold worker spawn.

## What

`SET duckgres.worker_ttl = '20m'` / `SHOW` / `RESET` now work mid-session, on all three protocol paths (simple, split-batch, extended):

- **Transpiler** intercepts the GUC like `duckgres.s3_cache` — validated with 22023 on a bad duration (without echoing client input), never forwarded to DuckDB, never intercepted inside a multi-statement batch.
- **Connection layer** applies it through a control-plane hook that updates the bound worker's pool-side hot-idle TTL; the next hot → hot_idle park persists the overridden `ttl_minutes`, which is exactly what both reapers (leader janitor, per-CP fallback) read. Session state flips only after a successful apply, so SHOW never reports a TTL the worker won't park with; a clamped value is stored clamped.
- **Same trust boundary as the startup option**: gated on `DUCKGRES_K8S_ALLOW_CLIENT_WORKER_PROFILE` (SET rejected with 22023 when off) and clamped to `DUCKGRES_K8S_WORKER_MAX_TTL`.
- **Exploratory tier**: SET binds a worker first on lazily-activated connections (the TTL is worker-side pool state), and the override is re-applied across tier worker switches; a failed re-apply fails the statement and resets session state, never the connection.
- **Standalone/process backends** accept SET/SHOW as session state only (no per-worker hot-idle TTL exists there).

`SHOW duckgres.worker_ttl` reports the session override, else the bound worker's current park TTL, else the connect-time baseline (startup GUC > org default > deployment default > built-in 1m).

## Notes / limitations

- TTL is stamped with whole-minute precision (pre-existing: `ttl_minutes` truncates), so sub-minute overrides round down — documented in the README.
- A CP restart between SET and disconnect loses the in-memory override (the park then uses the last persisted TTL); acceptable because the session dies with the CP connection anyway.

## Test plan

- `transpiler/worker_ttl_test.go`: SET/SHOW/RESET interception, normalization, 22023 rejections (no echo), multi-statement guard.
- `server/worker_ttl_test.go`: simple/batch/extended apply paths, lazy activation on SET, failure-keeps-state, gate 22023 passthrough, clamp reporting, escalation re-apply + failure semantics, state-only without the capability.
- `controlplane/worker_ttl_test.go` + `k8s_pool_worker_ttl_test.go`: pid→worker routing, gate/clamp hook, baseline resolution, pool SetWorkerTTL/WorkerTTL, and override persisted as `ttl_minutes` at park.
- `just lint` clean; full unit suites pass in both build flavors (`-tags kubernetes` and without). `controlplane/admin` Postgres-container tests fail in this environment (missing docker-compose) — pre-existing on main, unrelated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -378,13 +378,16 @@ connect) and **escalation** (small → standard, one-way). Env-only knobs:
   answers itself — `SET`/`SHOW duckgres.query_source`, ignored SETs, no-ops,
   `pg_stat_activity`, the empty query — MUST NOT acquire. That is the point of
   the whole feature; adding an acquire to an engine-free path silently deletes
-  the benefit. The `duckgres.s3_cache` GUC is the exception, on all three
-  protocol paths, because unlike the other duckgres GUCs it is WORKER state (it
-  rebuilds the worker's `ducklake_s3` secret): **`SET` always acquires** — the
-  swap needs a worker to apply to — and **`SHOW` acquires only when
+  the benefit. The `duckgres.s3_cache` and `duckgres.worker_ttl` GUCs are the
+  exception, on all three protocol paths, because unlike the other duckgres
+  GUCs they are WORKER state (the s3_cache secret swap; the worker_ttl
+  pool-side hot-idle TTL): **`SET` always acquires** — the apply needs a
+  worker to land on — and **`SHOW duckgres.s3_cache` acquires only when
   `c.hasPendingS3Cache`**, i.e. when a connect-time option has not been applied
   yet and answering first would report a transport the session is about to
-  leave (see the s3_cache section below).
+  leave (see the s3_cache section below). `SHOW duckgres.worker_ttl` never
+  acquires: until a worker exists, the connect-time baseline is the truthful
+  answer (there is no pending worker-side TTL state).
 - **A pinning FIRST statement acquires the standard profile directly** (one
   acquire, `pinned=true` → `MarkConnectionPinned`), never small-then-escalate.
 - **The pin set is the state boundary, and every member is load-bearing:** DML,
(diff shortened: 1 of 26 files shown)


fuziontech:
Rebased onto current main (was 8 days stale) + review fixes pushed as a second commit:

1. **Portal-Describe gap**: `Describe('P')` of SET/SHOW `duckgres.worker_ttl` fell through to the generic path → NoData + a DuckDB probe that acquires a worker on lazy exploratory connections. Both Describe variants now answer from session state (`TestWorkerTTLDescribeShow`).
2. **Escalation re-apply ordering**: a failed s3_cache re-apply early-returned, skipping the worker_ttl re-apply while keeping the session override → the new worker would park at baseline with SHOW lying. Both re-applies now run unconditionally; failures reset their own session state and join into one statement error (`TestEscalationReapplyRunsBothOverrides`).
3. **Whole-minute validation**: zero/sub-minute values rejected with 22023 — `ttl_minutes=0` means "deployment default" at reap time, so `SET '30s'` previously parked the worker for the *default* while SHOW reported 30s, and the README's `0s retires the worker` claim was false.
4. **e2e**: `worker_ttl_guc` harness assertion (SET/SHOW/RESET round-trip + sub-minute rejection).

Also added the missing CLAUDE.md section pinning the invariants. Full suites green both build flavors; `just lint` clean.

PR #1068, opened 2026-08-13
fix(controlplane): isolate service credentials

benben:
## Problem

Concurrent backend jobs sharing one audit principal can rotate the same service credential and invalidate each other. Authenticated `svc_` identities can also be rejected during admission because they have no ordinary org-user row.

## Changes

- Make every mint create a fresh credential ID and one-time secret; treat `principal` as non-unique audit metadata.
- Keep refresh and revoke scoped to an explicit credential ID.
- Admit persisted service credentials against the org vCPU budget without an org-user or per-user limit.
- Preserve handshake-only expiry and revocation semantics during queueing, lazy activation, and worker switching.
- Exclude historical grants from hot auth snapshots while retaining normal audit history.
- Hard-delete grant history when an org is deleted and purge pre-existing orphan grants before reusing an org name.
- Start credential TTL after lifecycle-lock contention and document a Duckgres-first, PostHog-second rollout.

## Testing

Added regressions for concurrent same-principal minting, ID-scoped refresh, service-credential admission, unknown IDs, post-handshake expiry/revocation, org deletion/name reuse, orphan cleanup, and TTL lock waits.

Automated checks run:

- `just test-configstore-integration`
- `just test-controlplane-k8s`
- Focused control-plane, provisioning, admin, and Postgres service-credential tests
- `just lint`
- `git diff --check`

The broader process integration stage reached an existing local socket timeout in `TestControlPlaneBasic`; its package tests and all scoped regressions passed.

## Rollout

Deploy this Duckgres change to the whole control-plane fleet before deploying the paired PostHog caller change. Compatibility with the legacy `force_rotate` body is intentionally one-way.

## Agent context

Human-driven, agent-assisted implementation and adversarial review using Codex subagents. All test identifiers and fixtures are synthetic; no operational or customer data is included.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -831,14 +831,13 @@ project-scoped (root-shaped: unrestricted at the org level).
 
 **Caller contract:**
 `POST /api/v1/orgs/:id/service-credentials` with
-`{principal, ttl_seconds?, force_rotate?}`. `principal` is audit attribution
-(`"dagster:events-backfill"`) AND the reuse key: one live grant per
-`(org_id, principal)`. `ttl_seconds` is clamped to [1 min, 1 h]
+`{principal, ttl_seconds?}`. `principal` is audit attribution
+(`"dagster:events-backfill"`) only: every request creates an independent grant
+with a new `credential_id` and secret, even when principals are identical.
+`ttl_seconds` is clamped to [1 min, 1 h]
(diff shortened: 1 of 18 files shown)


benben:
Paired caller change: https://github.com/PostHog/posthog/pull/82674. Roll out this Duckgres PR to the full control-plane fleet first.

PR #1081, opened 2026-08-17
logging: flush OTLP batch on worker/CP drain os.Exit

fuziontech:
## Summary

`os.Exit` skips `defer loggingShutdown()`, so the last drain/shutdown log line never left the OTLP `BatchProcessor`.

- Add `cliboot.FlushLogging()` — idempotent, no-op when export is off, shares `sync.Once` with the `InitLogging` shutdown closure.
- Call it only on the real drain exits: worker success and timeout, CP SIGTERM (including upgrade-drain immediate exit), and `drainAfterUpgrade`.
- Do not wrap startup `os.Exit(1)` fatals.

**Stack (1/3)** → #1082 → #1083

## Test plan

- [x] `go test ./internal/cliboot/ -run TestFlushLoggingIdempotent`
- [ ] Reviewer: confirm the listed `os.Exit` sites match the drain paths you care about

### DIFF
--- a/duckdbservice/service.go
+++ b/duckdbservice/service.go
@@ -969,8 +969,8 @@ func Run(cfg ServiceConfig) {
 		if !svc.WaitForDrain(ctx) {
 			slog.Warn("DuckDB service drain timed out before shutdown.", "timeout", workerShutdownDrainTime)
 			cancel()
-			cliboot.FlushLogging()
 			svc.CloseAll()
+			cliboot.FlushLogging()
 			os.Exit(0)
 		}
 		cancel()
--- a/duckdbservice/service.go
+++ b/duckdbservice/service.go
@@ -25,6 +25,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/flight"
 	"github.com/apache/arrow-go/v18/arrow/flight/flightsql"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/posthog/duckgres/internal/cliboot"
 	"github.com/posthog/duckgres/server"
 	"github.com/posthog/duckgres/server/flightclient"
 	"github.com/posthog/duckgres/server/observe"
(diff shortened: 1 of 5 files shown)


duckdbservice/service.go
```diff
@@ -968,12 +969,14 @@ func Run(cfg ServiceConfig) {
 		if !svc.WaitForDrain(ctx) {
 			slog.Warn("DuckDB service drain timed out before shutdown.", "timeout", workerShutdownDrainTime)
 			cancel()
+			cliboot.FlushLogging()
```
fuziontech: **[suggestion]** On drain timeout, `FlushLogging()` runs before `svc.CloseAll()`. `FlushLogging` is `provider.Shutdown`, after which further `OnEmit` no-ops. `CloseAll` still logs (query-log sink stop, catalog detach failures) — the lines you want after a forced teardown. The success path already flushes after `Shutdown()`/`CloseAll()`.

**Suggestion:** Move `cliboot.FlushLogging()` to after `svc.CloseAll()`, matching the success path. If flush-first is deliberate (protect the timeout line against a hung `CloseAll`), say so in one line and consider `ForceFlush` before `CloseAll` plus `Shutdown` after.

fuziontech (commented):
## Summary

The change correctly addresses `os.Exit` skipping `defer loggingShutdown()` by installing a process-wide `FlushLogging` hook that shares `sync.Once` with the `InitLogging` shutdown closure. Export-off and never-initialized paths are true no-ops, `provider.Shutdown` cannot run twice from the drain/`defer` pair, and the listed worker/CP drain `os.Exit(0)` sites match the real SIGTERM and tableflip upgrade exits (startup `os.Exit(1)` fatals are correctly left alone). The only real gap is the worker drain-timeout path flushing *before* `CloseAll`, which drops the teardown logs that path is most likely to emit.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

duckdbservice/service.go
fuziontech: Addressed in c145af0b1774e04231026b38555c87537c3ae7b9: timeout path now flushes after CloseAll so teardown logs can still export.

PR #1082, opened 2026-08-17
logging: redact SQL on OTLP, split service.name, default PostHog level warn

fuziontech:
## Summary

Turns the existing process-level OTLP export into a multi-tenant PostHog Logs system (still CP-only until worker Secret plumbing).

- PostHog branch: level → INFO sample → drop (`Starting metrics server`) → redact-and-keep query → OTLP.
- Default PostHog level is **warn**; stderr stays at `DUCKGRES_LOG_LEVEL`.
- Query text default is **redacted** (`RedactForLog`+4096). Secret DDL is a placeholder; ordinary SELECT text still leaves. **Never** redact on `LINE 1:`.
- `service.name` is the process role (`duckgres-control-plane` / `duckgres-worker` / `duckgres-reshard` / `duckgres`). `DUCKGRES_IDENTIFIER` becomes `duckgres.deployment`. This is also a VictoriaTraces dashboard cut.
- `InitLogging(bi)` and `InitTracing(bi)` share one resource. Entrypoints set `DUCKGRES_MODE` first.

**Stack** #1081 → **(2/3)** → #1083

Depends on #1081. Design doc: `docs/design/posthog-logs-multitenant.md`.

## Test plan

- [x] `go test ./internal/cliboot/` (strip handler, service.name, level/sample/drop)
- [ ] Confirm VictoriaTraces / existing log views that filtered `service.name=duckgres` or `duckgres-<identifier>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,9 +465,12 @@ on.
 
 The two exporters can be enabled independently, and the distinction matters
 because they carry different data. These events are metadata only. Application
-logs are not: `logQuery` / `logQueryError` attach the statement, and
+logs are not: OTLP keeps `RedactForLog`+4096 SQL only on exported WARN/ERROR.
+User-class `Query execution failed.` stays Info and does **not** export at the
+default WARN — those statements stay on stderr/`query_log` unless the PostHog
+level is raised. `logQuery` / `logQueryError` attach the statement, and
 `usersecrets.RedactForLog` only rewrites secret DDL, so ordinary SQL and its
-literals reach PostHog Logs.
+literals reach PostHog Logs only when that record is actually exported.
 
 | Set | Analytics events | Log export |
 | --- | --- | --- |
--- a/README.md
+++ b/README.md
@@ -407,11 +407,14 @@ Run with config file:
 | `DUCKGRES_EXPLORATORY_WORKER_CPU` | CPU request/limit of the exploratory worker pod (e.g. `1`, `500m`). Required (with the memory knob) for the tier to activate; a missing or invalid value logs a warning and leaves the tier OFF. Env-only. | - |
 | `DUCKGRES_EXPLORATORY_WORKER_MEMORY` | Memory request/limit of the exploratory worker pod (e.g. `2Gi`). Same requirement as the CPU knob. Env-only. | - |
 | `DUCKGRES_EXPLORATORY_WORKER_TTL` | Hot-idle TTL of exploratory worker pods (Go duration) — how long one stays parked for the org's next connection after its last one ends. Env-only. | `48h` |
-| `POSTHOG_API_KEY` | PostHog project API key (`phc_...`); enables log export **and product-analytics events**. Application logs carry query text — to get events without exporting SQL, leave this unset and use `POSTHOG_ANALYTICS_API_KEY` | - |
-| `POSTHOG_ANALYTICS_API_KEY` | PostHog project API key for product-analytics events **only**, leaving log export off. Takes precedence over `POSTHOG_API_KEY` for analytics | - |
+| `POSTHOG_API_KEY` | PostHog project API key (`phc_...`); enables log export **and product-analytics events**. Exported WARN/ERROR logs carry `RedactForLog`+4096 SQL (secret DDL is a placeholder). To get events without exporting SQL, leave this unset and use `POSTHOG_ANALYTICS_API_KEY` | - |
+| `POSTHOG_ANALYTICS_API_KEY` | PostHog project API key for product-analytics events **only**, leaving log export off. Takes precedence over `POSTHOG_API_KEY` for analytics. Never copied to worker pods | - |
 | `POSTHOG_HOST` | PostHog ingest host (shared by both exporters) | `us.i.posthog.com` |
-| `ADDITIONAL_POSTHOG_API_KEYS` | **(Experimental)** Comma-separated list of additional PostHog API keys to publish logs to. Requires `POSTHOG_API_KEY` to be set. | - |
-| `DUCKGRES_IDENTIFIER` | Suffix appended to the OTel `service.name` (e.g., `duckgres-acme`). Applies to **both** the log export and the OTLP trace export — they share one resource — so setting it renames the service in traces too, not just logs | - |
+| `ADDITIONAL_POSTHOG_API_KEYS` | **(Experimental)** Comma-separated extra PostHog API keys for log export. Requires `POSTHOG_API_KEY`. CP-only; not forwarded to workers | - |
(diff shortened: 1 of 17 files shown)


internal/cliboot/posthog_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	if l < slog.LevelInfo {
+		return h.inner.Enabled(ctx, l)
+	}
+	if h.sample <= 0 {
+		return false
+	}
+	if h.sample >= 1 {
+		return h.inner.Enabled(ctx, l)
+	}
+	return rand.Float64() < h.sample && h.inner.Enabled(ctx, l)
```
fuziontech: **[bug]** `infoSampleHandler.Enabled` decides INFO keep/drop with `rand.Float64() < h.sample`. The stack re-checks `Enabled` on the way in (`multiHandler.Handle`, `posthogLevelHandler.Handle`, `infoSampleHandler.Handle`). Each check is an independent Bernoulli trial, so `DUCKGRES_POSTHOG_LOG_INFO_SAMPLE=0.1` keeps about `0.1³` of INFO records. Defaults (`sample=0`, level warn) hide this; the knob is wrong the moment someone enables INFO sampling.

**Suggestion:** Make `Enabled` a pure level/config check (`sample > 0` for INFO). Perform the single random (or `trace_id` hash) draw only in `Handle`, and do not call `Enabled` again from the Handle wrappers. Add a test that a mid-range sample is not cubed.

internal/cliboot/otel_resource.go
```diff
@@ -3,20 +3,52 @@ package cliboot
...
+	if pod := os.Getenv("POD_NAME"); pod != "" {
+		attrs = append(attrs, semconv.ServiceInstanceID(pod), semconv.K8SPodName(pod))
+	}
+	if node := os.Getenv("NODE_NAME"); node != "" {
+		attrs = append(attrs, semconv.K8SNodeName(node))
+	}
+	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
+		attrs = append(attrs, semconv.K8SNamespaceName(ns))
+	}
+	return resource.NewWithAttributes(semconv.SchemaURL, attrs...)
```
fuziontech, line 40: **[suggestion]** Decision 10 treats the resource schema as a contract and requires `duckgres.mode` = raw `DUCKGRES_MODE`. `otelResource` never sets it, and `TestOTelServiceNameFromMode` does not assert it.

**Suggestion:** Append `attribute.String("duckgres.mode", os.Getenv("DUCKGRES_MODE"))` when the env is non-empty, and assert the attr in `TestOTelServiceNameFromMode`.

README.md
```diff
@@ -447,6 +454,8 @@ export POSTHOG_HOST=eu.i.posthog.com
 ./duckgres
 ```
 
+Remote worker pods do **not** inherit `POSTHOG_*` from the control-plane Deployment automatically — spawn builds an explicit env list. Until worker Secret-ref plumbing is in place, only processes that already have `POSTHOG_API_KEY` in their own env (control plane, process-backend children) export.
+
 ### PostHog Product-Analytics Events
 
 `POSTHOG_API_KEY` (and `POSTHOG_HOST`) also enables product-analytics event
```
fuziontech, line 461: **[suggestion]** The Product-Analytics Events paragraph still says `logQuery`/`logQueryError` attach the statement so ordinary SQL reaches PostHog Logs, with no level qualifier. At the default WARN, those Info lines do not export.

**Suggestion:** Qualify: OTLP keeps `RedactForLog`+4096 SQL only on exported WARN/ERROR; user-class failures stay on stderr/`query_log` unless the PostHog level is raised.

internal/cliboot/logging.go
```diff
@@ -301,7 +308,7 @@ func FlushLogging() {
 // Logs always go to stderr; PostHog is additive.
 // The log level is controlled by DUCKGRES_LOG_LEVEL (debug, info, warn, error).
```
fuziontech: **[nit]** `InitLogging` still documents that the log level is controlled only by `DUCKGRES_LOG_LEVEL`. After this PR, stderr and PostHog have independent levels.

**Suggestion:** Mention the PostHog level split in the doc comment.

fuziontech (commented):
## Summary

This PR lands the locked PR-2 handler stack correctly: `QueryStripHandler` sits only on the PostHog branch, snapshots query text (including a `WithAttrs` `origQuery` stash) before `RedactErrorForLog`, keeps `RedactForLog`+4096 query attrs, never treats `LINE 1:` as a redaction trigger, and leaves user-class `Query execution failed.` at Info so the default WARN export does not ship it. `InitLogging(bi)` and `InitTracing(bi)` share `otelResource(bi)`, every entrypoint sets `DUCKGRES_MODE` first, and `DUCKGRES_IDENTIFIER` is `duckgres.deployment` rather than a `service.name` suffix. The only real correctness defect is the INFO sampler: `Enabled()` rolls a new random draw and both the level wrapper and the sampler re-check `Enabled` on `Handle`, so a configured keep-fraction is applied about three times.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 1

internal/cliboot/posthog_handler.go
fuziontech: Addressed in 686796da8e025b50169cf09dc31bae0fd3a48e45: Enabled is now a deterministic sample>0 gate and the keep-draw happens once in Handle.

internal/cliboot/otel_resource.go
fuziontech: Addressed in 686796da8e025b50169cf09dc31bae0fd3a48e45: otelResource now sets duckgres.mode from DUCKGRES_MODE and TestOTelServiceNameFromMode asserts it.

README.md
fuziontech: Addressed in 686796da8e025b50169cf09dc31bae0fd3a48e45: Product-Analytics paragraph now says RedactForLog+4096 SQL exports only on WARN/ERROR, not user-class Info failures.

internal/cliboot/logging.go
fuziontech: Addressed in 686796da8e025b50169cf09dc31bae0fd3a48e45: InitLogging docs now split DUCKGRES_LOG_LEVEL (stderr) from DUCKGRES_POSTHOG_LOG_LEVEL (default warn).

fuziontech:
Superseded after #1081 squash: the original stack history conflicts with the squash commit. Replacement PR targeting main is the rebased branch — closing this one once the replacement is up.

fuziontech:
Superseded by #1088 (rebased onto main after #1081 squash).

PR #1083, opened 2026-08-17
logging: stamp pid on CP loggers; session-scope user/pid on worker WARNs

fuziontech:
## Summary

Makes CP and worker logs joinable by the existing backend-key `pid` without leaking the previous user on hot-idle worker reuse.

- `clientConn.logger()` and `handleConnection` `clog` stamp `pid`.
- Workers attach `user`+`pid` on a session logger and clear it on `DestroySession`. Org/worker stay on the process default logger — never `SetDefault(user)`.
- Stuck-query WARN (`Query appears stuck — no progress detected.`) uses the session logger.
- `WorkerCreateSessionPayload` carries optional `pid`.

**Stack** #1081 → #1082 → **(3/3)**

Depends on #1082.

## Test plan

- [x] `go test ./duckdbservice/ ./server/ ./controlplane/`
- [x] `TestStuckQueryWarnCarriesSessionIdentity`, `TestNeverSetDefaultWithUser`, `TestClientConnLoggerIncludesPID`

### DIFF
--- a/duckdbservice/service.go
+++ b/duckdbservice/service.go
@@ -973,8 +973,9 @@ func Run(cfg ServiceConfig) {
 		if !svc.WaitForDrain(ctx) {
 			slog.Warn("DuckDB service drain timed out before shutdown.", "timeout", workerShutdownDrainTime)
 			cancel()
-			cliboot.FlushLogging()
+			// Close first so teardown WARNs land in the batch, then flush.
 			svc.CloseAll()
+			cliboot.FlushLogging()
 			os.Exit(0)
 		}
 		cancel()
@@ -1325,7 +1326,11 @@ func (p *SessionPool) DestroySession(token string) error {
 	p.mu.Lock()
 	session, ok := p.sessions[token]
 	stop := p.stopRefresh[token]
+	var log *slog.Logger
 	if ok {
+		// Snapshot before clear so destroy-path WARNs keep user+pid
+		// without leaving the field set for a later Logger() leak.
+		log = session.Logger()
 		delete(p.sessions, token)
 		delete(p.stopRefresh, token)
 		clearSessionLog(session)
@@ -1437,7 +1442,7 @@ func (p *SessionPool) DestroySession(token string) error {
 	if p.sharedWarmMode && p.maxSessions == 1 && session.DB != nil {
 		wipeCtx, wipeCancel := context.WithTimeout(context.Background(), userSecretOpTimeout)
 		if _, err := wipeUserSecrets(wipeCtx, session.DB); err != nil {
-			slog.Warn("Failed to wipe user secrets on session destroy.", "user", session.Username, "error", err)
+			log.Warn("Failed to wipe user secrets on session destroy.", "user", session.Username, "error", err)
 		}
 		wipeCancel()
 	}
@@ -1448,7 +1453,7 @@ func (p *SessionPool) DestroySession(token string) error {
 	// which the worker's checkpointer would write past the cache proxy.
 	if p.sharedWarmMode {
 		if err := p.SetS3CacheEnabled(true); err != nil {
-			slog.Warn("Failed to restore S3 cache transport on session destroy.", "user", session.Username, "error", err)
+			log.Warn("Failed to restore S3 cache transport on session destroy.", "user", session.Username, "error", err)
 		}
 	}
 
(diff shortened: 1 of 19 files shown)


duckdbservice/service.go
```diff
@@ -1324,6 +1328,7 @@ func (p *SessionPool) DestroySession(token string) error {
 	if ok {
 		delete(p.sessions, token)
 		delete(p.stopRefresh, token)
+		clearSessionLog(session)
```
fuziontech, line 1336: **[suggestion]** `clearSessionLog` runs here, before the destroy-path WARNs. `Failed to wipe user secrets on session destroy` and `Failed to restore S3 cache transport on session destroy` then use `slog.Warn` with `user` but no `pid`.

**Suggestion:** Snapshot `log := session.Logger()` before `clearSessionLog` so wipe/restore failures keep `user`+`pid`, then nil the field.

server/conn.go
```diff
@@ -593,6 +593,12 @@ func (c *clientConn) logger() *slog.Logger {
 	if c.workerPod != "" {
 		attrs = append(attrs, "worker_pod", c.workerPod)
 	}
+	if c.pid != 0 {
+		attrs = append(attrs, "pid", c.pid)
+	}
+	if qs := c.QuerySource(); qs != "" && qs != "standard" {
```
fuziontech: **[suggestion]** The pid stamp on `clientConn.logger()` is correct. This hunk also stamps `query_source` whenever it is non-`standard`. `logClientQueryReceived` already adds `query_source` (duplicate key), and `ConnectionBilling` deliberately logs only `value_len` on a non-canonical source — `logger()` will now attach the raw `QuerySource()` value on that Warn.

**Suggestion:** Drop `query_source` from `logger()` in this PR, or keep it and stop repeating it at `logClientQueryReceived`. If it stays, clamp or omit it on the non-canonical billing path.

fuziontech:
Also addressed the out-of-diff finding: `transient.go` retry helpers now take `*slog.Logger` and `flight_handler` passes `session.Logger()`. Nil falls back to `slog.Default()` only when no session exists. `TestRetryWarnCarriesSessionIdentity` asserts user+pid on a session-bound retry WARN. Landed in b6764cbed132cad40f127dbc1aa1fe0a6489e91a.

fuziontech (commented):
## Summary

The load-bearing contract is implemented correctly: org/worker stay on `slog.Default()` via `workerIdentityAttrs` (never `SetDefault(user)`), the session logger is attached with `user`+`pid` and cleared in `DestroySession`, stuck-query WARNs go through `logStuckQuery` → `Session.Logger()`, and the CP backend-key pid is plumbed `ReservePID` → `clog.With("pid")` / `clientConn.logger()` → `ManagedWorker.CreateSession` → `WorkerCreateSessionPayload.PID` → `attachSessionLog`. `attachSessionLog` after a failed `CreateSession` is safe because every error path returns a nil `*Session` and the helper no-ops. Remaining worker WARNs on the live-query retry path and the destroy wipe/restore path still log through `slog.Default()` (user via an explicit attr at best, never pid), so those lines cannot join to the CP connection.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 0

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[suggestion]** duckdbservice/transient.go:119 -- Live-query retry helpers still emit `slog.Warn` with no session identity. `retryOnTransient` / `recoverAbortedTransaction` / `retryOnConflict` are called from `flight_handler` with a live `*Session`. Those WARNs cannot join to a CP connection.
  - **Suggestion:** Thread `session.Logger()` into the three retry helpers. Keep `slog.Default()` only when no session exists.

duckdbservice/service.go
fuziontech: Addressed in b6764cbed132cad40f127dbc1aa1fe0a6489e91a: snapshot log := session.Logger() before clearSessionLog so wipe/restore WARNs keep user+pid.

server/conn.go
fuziontech: Addressed in b6764cbed132cad40f127dbc1aa1fe0a6489e91a: dropped query_source from clientConn.logger(). logClientQueryReceived still attaches it; ConnectionBilling no longer inherits a raw client value.

fuziontech:
Superseded by the rebased PR targeting main after #1081/#1088 landed.

PR #1084, opened 2026-08-17
k8s: copy PostHog log env as secretKeyRef; never fail spawn

fuziontech:
## Summary

Worker and reshard pods get the same PostHog log export config as the control plane, without putting the project token in the pod spec as a plaintext `value:` — and without failing spawn if the logging env is missing.

- Cache allowlisted named env from `Get(namespace, POD_NAME)` once at pool start. `SecretKeyRef` is resolved when the **worker** pod starts.
- `POSTHOG_API_KEY` is copied only as `valueFrom.secretKeyRef`. A literal `value:` is refused. `envFrom` is not invented-around (unit test: CP pod with only `envFrom` → worker has no key).
- Missing `POD_NAME`, Get failure, or missing named env: one WARN (`PostHog log env not found on CP pod spec; workers will not export.`) and omit. **Never fail spawn.**
- Allowlist: `POSTHOG_API_KEY`, `POSTHOG_HOST`, `DUCKGRES_POSTHOG_LOG_*`, `DUCKGRES_IDENTIFIER`. Do **not** forward `ADDITIONAL_POSTHOG_API_KEYS` or `POSTHOG_ANALYTICS_API_KEY`.
- Reshard allowlist uses the same copy helper. Inject Downward API `POD_NAMESPACE` next to `POD_NAME`/`NODE_NAME`.
- Health-check JSON grows `otlp_export_enabled` + `otlp_export_failures` (process-lifetime monotonic). CP stores last-seen per worker and `Add(delta)` only (restart → treat last as 0 then `Add(n)`). Metric: `duckgres_otlp_log_export_failures_total{source,reason}` — **no `{org}`**. Workers do not call `InitMetrics`.

**Stack** #1081 → #1082 → #1083 → **(4/5)**

Depends on #1083. Design: `docs/design/posthog-logs-multitenant.md` (Key Decisions 5/12).

## Test plan

- [x] `go test -tags kubernetes ./controlplane` — `TestWorkerSpawnPostHogEnvIsSecretRef`, `TestWorkerSpawnOmitsPostHogWhenCPHasOnlyEnvFrom`, `TestWorkerSpawnSucceedsWhenCPPodGetFails`, `TestReshardAllowlistIncludesPostHog`
- [x] `go test ./controlplane ./duckdbservice ./internal/cliboot` — health JSON fields + last-seen delta roll-up
- [ ] Charts follow-up: named `POSTHOG_API_KEY` `secretKeyRef` on the CP Deployment (not mergeable here)

PR #1085, opened 2026-08-17
docs/e2e: PostHog Logs multi-tenant contract and worker env asserts

fuziontech:
## Summary

Closes the user-visible contract for PostHog Logs multi-tenant now that worker Secret-ref plumbing exists (#1084).

- README: worker/reshard named `secretKeyRef` copy, never-fail-spawn, charts contract, CP-only first slice, egress proof.
- CLAUDE.md: load-bearing contract (`service.name` roles, redacted query default, WARN default, named secretKeyRef, never fail spawn, `pid` join key, `FlushLogging` drain sites, do not replace `query_log`).
- `assert_worker_pod`: worker must **not** have a plaintext `POSTHOG_API_KEY` `value:`. If the CP named env has a `secretKeyRef`, the worker has the same name+key. Until charts set that, the copy branch is skipped; the plaintext assert still runs. Do **not** require a successful export.
- `tests/mw-dev/README.md`: ingest cannot be asserted in-Job (the Job cannot read PostHog — same reason as analytics). First mw-dev `duckgres-worker` record in the **analytics** project is the egress proof.
- Manifests: keep asserting 443 egress on worker + CP + reshard.

**Stack** #1081 → #1082 → #1083 → #1084 → **(5/5)**

Depends on #1084.

## Test plan

- [x] `go test ./tests/manifests/`
- [x] `bash -n tests/mw-dev/e2e/harness.sh`
- [ ] Charts follow-up still required before the secretKeyRef copy branch fires on mw-dev

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -495,6 +495,43 @@ release lets shutdown kill live work); `reapIdle` releases tokens stranded by a
 `GetFlightInfo` whose `DoGet` never arrived. `terminationGracePeriodSeconds=3600`
 (`k8s_pool.go`) must stay above `workerShutdownDrainTime` (55m).
 
+## PostHog Logs (OTLP) — LOAD-BEARING CONTRACT
+
+Process-level slog → PostHog Logs via OTLP (`internal/cliboot.InitLogging`).
+This is **not** a replacement for `ducklake.system.query_log`. Product-analytics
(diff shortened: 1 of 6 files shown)


fuziontech:
Superseded by a rebase onto current `main` (after #1091). Closing this stacked PR in favor of the replacement.

PR #1114, opened 2026-08-21
Provision multi-tenant Trino cells

fuziontech:
Makes Trino a query engine the control plane provisions, so a customer enabled for Trino gets a catalog, credentials, authorization and resource limits on a shared cluster without anyone editing the cluster.

This restores the customer-facing Trino subsystem removed in `6926069` (a product decision, not a technical failure) and ports it to DuckLake, since Iceberg/Lakekeeper no longer exists.

## What a cell gets per enabled org

| Artifact | Where |
|---|---|
| catalog `org_<id>` | `CREATE CATALOG` over REST, persisted by the PostgreSQL catalog store |
| `password.db` / `group.db` | Secret `trino-auth`, coordinator only |
| metadata-database password | Secret `trino-tenant-secrets`, one key per org, mounted as a file |
| OPA bundle | pulled by the coordinator's sidecar |
| resource group | ConfigMap `trino-resource-groups` |

## Two things that changed for the better

**No catalog carries a password.** Trino logs the `CREATE CATALOG` statement and shows it in the web UI, and sends catalog properties to workers. A catalog therefore names a file — `ducklake.metadata.connection-password-file` — and the control plane writes the password into a Secret. A test asserts the property map contains no password and that `ducklake.metadata.connection-password` is never set.

**The IAM bypass is gone.** The old provisioner built each catalog's role ARN from an env-configured AWS account, with `allowNoIamRole` to skip it. That switch could put every tenant catalog on the Trino pod's own identity, which can assume *any* `duckling-*` role — collapsing the per-org S3 boundary. The ARN now comes from the warehouse record, and a warehouse without one stays `pending` rather than being provisioned without a boundary.

## Tenant isolation

Three things, and only these, separate customers: the OPA policy, `s3.iam-role` on each catalog, and the per-catalog metadata credentials.

This also closes a leak the previous iteration had: the policy allowed `ExecuteQuery` unconditionally and never restricted query visibility, so one org could read another's **SQL text** through `system.runtime.queries` and the web UI, and kill its queries. `ViewQueryOwnedBy`, `FilterViewQueryOwnedBy` and `KillQueryOwnedBy` are now same-org only, derived from the existing `group_catalogs` map rather than a second source of truth. `__admin_provisioner` gets no cross-tenant visibility — the reconcile loop only issues `SHOW`/`CREATE`/`DROP CATALOG`, so the grant would buy nothing and expose everything.

Worth knowing for anyone editing the policy: the filter operation is `FilterViewQueryOwnedBy`, not `FilterViewQuery`. A rule on the wrong name matches nothing and denies silently while looking correct, so there is a static test on the operation names. Related, the comment claiming `opa.allow-permission-management-operations` keeps these operations away from OPA was wrong — it only guards grant/deny/revoke — which is why the gap was live rather than theoretical.

## Scope

One cell (`DUCKGRES_TRINO_CELL_ID`, default `cell-001`), structured so a per-cell map replaces the single coordinator URL later. No assignment policy, capacity management or rebalancing. `AlterCatalog` is wired but unexercised. `opa.policy.batched-uri` must stay off — the batched input shape fails every filter rule closed, and a test locks it.

## Verification

`go build`/`go vet` clean, `golangci-lint` no new issues, `go test -p 1 -tags kubernetes ./controlplane/...` green, config-store tests against real PostgreSQL green. Both the projection gate and the query-visibility rules were verified to fail when deliberately broken.

Needs the chart side: PostHog/charts `trino-cells` mounts `trino-tenant-secrets` at `/etc/trino/tenant-secrets` on coordinator **and** workers, and restores the cross-namespace RBAC.


fuziontech:
Rebased on `main`, and one thing changed that is worth calling out.

**Migration renumbered 037 → 038.** `000037_add_org_max_hot_idle_workers` landed on main while this branch was open, so both files claimed 037. Goose records a migration by its version number, so whichever ran second would have been recorded as already applied and its tables never created — the Trino tables would simply not exist, and the provisioner would fail on first reconcile. Now `000038_add_trino_tables.sql`, with the version assertions in `migrations_postgres_test.go` updated to match.

The only conflict was in that same test file (both sides editing the downgrade blocks); resolved by keeping both sets of DROPs.

Verified after the rebase: `go build`/`go vet -tags kubernetes` clean, and `./controlplane/provisioner/...` (the provisioner and OPA policy suites) green uncached.

One note for anyone running the full suite locally: `./controlplane/admin/...` and the other Postgres-backed packages fail here with `column "max_hot_idle_workers" of relation "duckgres_orgs" does not exist`. That is **not** from this branch — it reproduces identically on a clean `origin/main` worktree, and is a stale local test database missing main's own migration 037. CI, which starts from a fresh database, is green.

PR #1116, opened 2026-08-21
Identify Trino tenants by database_name, not org id

fuziontech:
Follows https://github.com/PostHog/duckgres/pull/1114. **Worth landing before the first org is enabled** — no catalogs exist yet, so there is nothing to migrate. After a tenant is live this becomes a rename with a customer-visible break.

## The problem

The Trino principal was the org id, which `validateDucklingOrgID` allows to be a canonical UUID. For PostHog's own org that means:

```
host=trino.dw.us.postwh.com  user=4dc8564dbd8210652f4097f7c50f67cf
SELECT * FROM org_4dc8564dbd8210652f4097f7c50f67cf.main.events
```

Nothing else in duckgres asks a customer to know their org id. For the DuckDB warehouse the public identity is `database_name` — the comment on `databaseNamePattern` says it outright: *"the tenant's public socket identity — the single hostname label of its managed SNI hostname and the dbname clients connect with."* Trino was the only surface leaking the internal id.

## The change

The principal is now `database_name`, so one name identifies a tenant across both engines:

```
host=trino.dw.us.postwh.com  user=acme-analytics
SELECT * FROM org_acme_analytics.main.events
```

Same password as before — `password.db` still carries the `duckgres_org_users` root bcrypt hash, the same one the DuckDB warehouse authenticates with.

**Safe as a catalog stem** because `duckgres_orgs.database_name` has a global unique index (`idx_duckgres_orgs_database_name`), so two tenants cannot derive the same catalog name.

**The `org_` prefix stays.** It is what `opa.ManagedCatalogPattern` (`^org_[a-z0-9_]+$`) anchors on, and that pattern is the guard that stops the provisioner from ever dropping `system`, `jmx`, or a hand-rolled catalog. Dropping the prefix would mean matching bare labels, which weakens a boundary I would rather leave alone.

## Scope: only names the customer types

| Derived from | What |
|---|---|
| `database_name` (new) | Trino username, catalog name, group label, resource-group path |
| `org_id` (unchanged) | config store PK, tenant-password Secret key, metadata database, duckling IAM role, S3 path |

The internal plumbing is deliberately untouched — visible in the test fixtures, where a catalog for org `42` is now `org_db42` but still points at `mdstore_42`, `/etc/trino/tenant-secrets/42` and `role/duckling-42`.

## Orgs without a database_name

They have no derivable identity, so `ListTrinoEnabledOrgs` now drops them — the same treatment an org with no root user already gets, and for the same stated reason. Projecting a catalog named `org_` would collide across every such org. The join moved from `LEFT` to `INNER` plus a non-blank check, with a defensive guard in the catalog loop too.

## Renames

`TestAdminUpdateOrgRenamesDatabaseNamePostgres` shows `database_name` is mutable. The existing reconcile handles it without new code: the old catalog falls out of the wanted set and is dropped, the new one is created, and both match the managed pattern so the drop is in-bounds. The tenant's connection breaks — as renaming already breaks its DuckDB hostname, so the behaviour is consistent rather than new.

## Tests

- New `TestTrinoIdentityIsDatabaseNameNotOrgID` pins the whole contract end to end and asserts the org id appears in **neither** auth file nor `resource-groups.json`. It uses a hyphenated name so the sanitize split is covered: group `org_acme_analytics`, selector user `acme-analytics` (raw, because it matches Trino's `current_user`).
- New `TestBuildTrinoAuthFiles_SkipsOrgWithoutDatabaseName` and `TestListTrinoEnabledOrgsRequiresADatabaseName`.
- Every existing fixture now carries a `DatabaseName` **distinct** from its org id (`42` → `db42`), so the suite fails if the principal ever reverts.
- `go test -tags kubernetes ./controlplane/provisioner/... ./tests/configstore/...` green.

Unrelated pre-existing failures in `controlplane/admin` (`column "max_hot_idle_workers" does not exist`) reproduce on a clean `main` — stale local test schema, not this change.

### DIFF
--- a/controlplane/configstore/models.go
+++ b/controlplane/configstore/models.go
@@ -524,6 +524,26 @@ type TrinoEnabledOrg struct {
 	State            ManagedWarehouseProvisioningState // current state at read time
 }
 
+// TrinoPrincipal is the tenant's customer-facing identity in Trino: the
+// username it authenticates as, and the stem every derived name is built
+// from (catalog, group, resource group).
+//
+// It is DatabaseName, not OrgID, so a tenant is known by the same name in
+// Trino as in its DuckDB warehouse — where database_name is already the
+// public socket identity (the SNI hostname label and the dbname clients
+// connect with). OrgID may be a bare UUID, which would put a hex string in
+// the customer's connection string and in every fully-qualified table
+// reference.
+//
+// Safe as a catalog stem because duckgres_orgs.database_name carries a
+// global unique index, so two tenants cannot derive the same catalog name.
+// OrgID remains the primary key and keys all internal plumbing (config
+// store rows, the tenant-password Secret); only names the customer types
+// are derived from this.
+func (o TrinoEnabledOrg) TrinoPrincipal() string {
+	return o.DatabaseName
+}
+
 // NOTE: the cluster-wide singleton config tables (global_config,
 // ducklake_config, rate_limit_config, query_log_config) were removed — they
 // were seeded and served by the admin API but never read to drive runtime
--- a/controlplane/configstore/trino.go
+++ b/controlplane/configstore/trino.go
@@ -197,25 +197,33 @@ func (cs *ConfigStore) DisableTrino(orgID string) error {
 // runs in the same handler that toggles Enabled), and silently skipping is
 // safer than projecting a half-built password file.
 //
+// Orgs with no duckgres_orgs row, or a blank database_name, are skipped for
+// the same reason: database_name is the org's Trino principal (see
+// TrinoEnabledOrg.TrinoPrincipal), so without it there is no username to put
+// in password.db and no stem to build a catalog name from. Skipping leaves
+// the org Pending rather than projecting a catalog called `org_`.
+//
 // Every cell's rows are returned, unassigned ones included; the caller
 // filters by cell (see TrinoEnabledOrg.CellID for why the filter is not in
 // the SQL).
 func (cs *ConfigStore) ListTrinoEnabledOrgs() ([]TrinoEnabledOrg, error) {
 	var out []TrinoEnabledOrg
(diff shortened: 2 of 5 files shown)


fuziontech:
Self-review before landing turned up a cross-tenant hazard in this PR, now fixed in `b28be35`.

Sanitization is injective over principals that satisfy `ValidateDatabaseName` — that grammar is lowercase alphanumerics and hyphens, the hyphen is the only character sanitization rewrites, and no valid principal contains the underscore it becomes. But **grandfathered rows predate that validation** (the `ValidateDatabaseName` doc comment says so explicitly), so a stored `database_name` can hold an underscore or uppercase — and then `Acme_Corp` and `acme-corp` both derive catalog `org_acme_corp`.

That is not a naming annoyance, it is a cross-tenant read:

1. First org creates `org_acme_corp` against its own metadata store.
2. Second org finds the name already present → recorded `Existed` → marked **Ready**.
3. The OPA bundle grants group `org_acme_corp` the catalog `org_acme_corp`.
4. Second org queries the first org's data.

`rejectPrincipalCollisions` now holds **both** orgs back and reports why. Refusing to provision either is the only safe resolution — picking a winner by sort order silently hands one tenant a catalog the other also believes it owns.

The check runs **before all five projection steps**, not inside the catalog step. That placement is the load-bearing part: a colliding org that still reached the OPA bundle would be granted the contested name, and the other org's legitimately created catalog answers to it — so filtering only at catalog-creation time would leave the hole open.

Also corrected the `TrinoCatalogName` doc comment, which claimed uniqueness merely "depends on the unique index holding post-sanitization". For valid names it is stronger than that (injective by construction); for grandfathered ones it is weaker, and neither was captured.

Tests: `TestRejectPrincipalCollisions` (both contested orgs held, uncontested one unaffected) and `TestRejectPrincipalCollisions_NoneWhenDistinct`.

PR #1118, opened 2026-08-22
Answer OPA's batched filtering shape

fuziontech:
Found while validating the first pilot org. **Any `information_schema` listing over a real catalog fails**, which breaks JDBC clients and BI tools on connect.

```
Failed to query OPA backend
RejectedExecutionException: Max requests queued per destination 1024 exceeded
  io.trino.plugin.opa.OpaHttpClient.parallelRequest
```

## Cause

With only `opa.policy.uri` set, the coordinator issues **one OPA request per candidate object**. PostHog's catalog has 57 schemas and well over 1024 tables, so `filterTables` overruns Jetty's per-destination queue.

OPA was not the bottleneck — it was answering in ~0.4ms with 200s. It had simply been asked 18,724 times.

## Why this wasn't already done

The policy carried an explicit DO-NOT-ENABLE warning, and it was right: `opa.policy.batched-uri` against a policy that only reads the singular `input.action.resource` denies **every** filter query. That is fail-closed and safe, but from the outside it is indistinguishable from a tenant-isolation bug. The warning asked for the policy to be extended first — this is that.

## Approach

`batch` is not a parallel set of rules. It re-evaluates `allow` with each candidate substituted in:

```rego
batch contains i if {
	some i
	resource := input.action.filterResources[i]
	allow with input.action.resource as resource
}
```

So batched and non-batched are the same decision by construction, and any rule added to `allow` is honoured in both. That property is what keeps this from becoming a second isolation surface that can silently drift from the first.

`FilterColumns` needs its own rule because its batched shape is not simply a list of the non-batched resource: Trino sends a **single** table candidate carrying a `columns` array and expects indices into *that* array. The columns are fanned out into one table resource each so the same `allow` rules answer for them.

## Tests

The old `TestBatchedInputDeniesByDefault` guard is gone — it locked in the behaviour this PR deliberately changes. Replacing it:

- `TestBatchedFilteringMatchesNonBatched` — for FilterCatalogs, FilterSchemas and FilterTables, asserts the batched result is exactly the allowed index set **and** re-checks every candidate through `allow`, failing if the two shapes disagree.
- `TestBatchedFilterColumns` — all three columns of the caller's own table are visible; **no** column of another org's table is.

## Rollout ordering (important)

The bundle is served by the control plane, so the policy must be deployed **before** the chart sets `opa.policy.batched-uri`. Flipping the chart first points the coordinator at `/v1/data/trino/batch` on a bundle that has no `batch` rule — undefined, empty result, everything filtered out.

Order: merge this → promote duckgres to prod → then the charts PR that sets the batched URI.

### DIFF
--- a/controlplane/provisioner/opa/policy.rego
+++ b/controlplane/provisioner/opa/policy.rego
@@ -86,20 +86,26 @@
 # so when per-user identity within an org lands (OIDC), org-mates keep
 # seeing each other's queries with no policy change.
 #
-# !!! Cross-component invariant: this policy is the NON-BATCHED OPA
-# contract. Trino 476's batched access control (OpaBatchAccessControl,
-# enabled via `opa.policy.batched-uri` on the trino-opa plugin) sends
-# candidates under `input.action.filterResources` (plural) instead of
-# `input.action.resource` (singular). Every filter rule below reads
-# `input.action.resource.*`, so under the batched shape they all fail
-# their body and fall through to `default allow := false`. That is
-# fail-closed — safe by the threat model — but operationally it means
-# enabling batched mode in the chart silently denies every filter
-# query, which looks like a tenant-isolation bug from the outside.
-# DO NOT set `opa.policy.batched-uri` in the customer-Trino chart
-# without first extending this policy to handle the batched shape (and
-# the bundle generator's tests). TestBatchedInputDeniesByDefault below
-# locks the fail-closed behaviour as a regression guard.
+# !!! Cross-component invariant: this policy serves BOTH OPA shapes.
+# `allow` answers the non-batched contract, where the candidate sits at
+# `input.action.resource` (singular). `batch` answers Trino's
+# OpaBatchAccessControl (`opa.policy.batched-uri`), where candidates
+# arrive as a list at `input.action.filterResources` (plural) and the
+# response is the list of allowed INDICES into it.
+#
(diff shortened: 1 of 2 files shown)


fuziontech:
Self-review before asking for a stamp. One real fix, in `9a8bffb`.

**Out-of-bounds index on FilterColumns.** `FilterColumns` returns indices into the candidate's `columns` array, not into `filterResources` — but the generic rule matched it too. For a readable table carrying **no** columns the generic rule contributed index `0`, an index into an empty array, which the coordinator maps back against a zero-length list.

Verified directly against the compiled policy:

```
before   ZERO-COLUMNS: ["0"]
after    ZERO-COLUMNS: []
```

The generic rule now skips `FilterColumns` and the dedicated rule owns it: one rule per shape. Worth noting this is a deliberate divergence from upstream's batch example, which has the same overlap.

Two other things I checked rather than assumed:

- **A fully-denied batch returns `[]`, not undefined.** That matters: `[]` means "filter everything out" (correct fail-closed), whereas undefined could surface as a backend error. Confirmed the partial set rule yields a defined empty array with no matches, and that no conflicting `default batch` exists in the policy.
- **`allow` is unchanged.** `batch` only re-evaluates it with a substituted resource, so this PR adds no new authorization surface — which is the reason for that construction rather than a parallel rule set.

Added a regression test for the column-less table alongside the existing cross-tenant assertion.

PR #1171, opened 2026-09-11
feat: support explicit registry-only Trino deployments

benben:
## Summary

- Add explicit `DUCKGRES_TRINO_REGISTRY_ONLY=true` support for deployments with registered cells and no legacy coordinator. The default still rejects a missing legacy URL; registry-only mode requires a valid registry and rejects mixed configuration.
- Require explicit initial placement before enablement in registry-only mode. Preserve existing ownership and legacy defaults; never assign an arbitrary registered cell or reinterpret legacy rows.
- Expose safe unassigned status and an explicit cell picker on operational Trino pages. No warehouse migration or Gateway routing changes.

## Validation

- Red/green startup, admin and enablement regressions.
- `just test-trino`, `just test-controlplane-k8s`, `just test-mw-fixtures`, `just ui-test` (195 tests), and `just lint` pass locally.
- Independent adversarial review completed.
- The first live run passed legacy compatibility, initial placement, tenant isolation and green hydration. Its registry-only phase found a fixture Service-convergence race after restart; bounded registry checks and behavioral regression tests now cover that transition.
- The isolated Trino CI lane now restarts its control plane in registry-only mode after existing legacy/multicell checks, verifies both registered backends, legacy-owner rejection, and absence of the legacy bundle route, then restores its fixture.

Draft until the extended live Trino CI lane passes. This does not alter a shared deployment.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -921,7 +921,7 @@ impersonation, audit log; sliceable by org + user). Design + decisions:
   Every read is cached + timeout-bounded and degrades to `available:false`
   plus a reason rather than erroring the page: the console must render
   during exactly the incident it exists for. Unset
-  `DUCKGRES_TRINO_COORDINATOR_URL` leaves the routes unregistered.
+  legacy URL and registry configuration leaves the routes unregistered.
 - Touching any of the above → update `controlplane/admin/*_test.go` (esp
   `authz_test.go`, `kill_switch_test.go`, `operators_api_test.go`,
   `trino_test.go`, `trino_client_test.go`),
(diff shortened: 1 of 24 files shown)


bill-ph (approved):
Reviewed the registry-only Trino deployment path, including strict configuration validation, explicit initial-cell admission before enablement, fail-closed handling for unassigned/unknown ownership, and the accompanying admin/E2E regression coverage. I found no P0 blocker requiring a code change. The PR is appropriately kept as draft pending the extended live Trino CI lane; current CI is still in progress.\n\n— Robo Bill

PR #1172, opened 2026-09-11
feat(admin): configure Trino at the top of org settings

benben:
## Summary

- Put Trino configuration first in the org configuration card, with separate **Select cell**, **Enable Trino**, and **Disable Trino** actions.
- Require an explicit saved initial assignment before enabling. Preserve existing assignments and resource-group tiers, including while disabled.
- Reuse the authenticated, admin-only, audited enable/disable endpoints. No provisioning token is exposed to the browser.
- Keep other unsaved org fields untouched by Trino actions. Retain the lower Trino card for provisioning status and live query counts.

This does not implement warehouse moves, maintenance barriers, or Gateway routing. Disabling Trino remains asynchronous reconciliation, not proof that all work has drained.

## Validation

- TDD: new top-of-configuration placement test failed before implementation, then passed.
- `just ui-test`: typecheck, 210 tests, and production build passed.
- `just lint`: passed with zero issues.
- Coverage includes read-only viewers, immutable assignments, saved selection before enable, preserved tier, pending/error states, ambiguous writes, unconfigured environments, registry-only unassigned recovery, request shapes, and unsaved general org fields.
- Independent agent review completed; no blocking findings. Added the review's registry-only regression cases.

UI-only change over existing backend endpoints. No live warehouse settings were changed during development.

### DIFF
--- a/controlplane/admin/README.md
+++ b/controlplane/admin/README.md
@@ -202,7 +202,7 @@ routes accept `?cell=<logical-id>` and default to `legacy`. Org detail resolves
 its authoritative stored assignment regardless of a supplied cell parameter.
 Each coordinator has separate caches; tenant counts include only its cell.
 
-Admins can select an initial cell in the org's Trino card or send
+Admins can select an initial cell at the top of **Org configuration** or send
 `PUT /api/v1/orgs/:id/trino/cell` with `{"cell":"cell-001"}`. This stores a
 disabled assignment before the normal enable endpoint is called. The warehouse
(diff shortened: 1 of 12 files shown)


bill-ph (approved):
Reviewed the admin UI changes for top-level Trino configuration, explicit initial cell selection, separate enable/disable actions, immutable assignments, viewer read-only behavior, and preservation of unsaved org fields. The UI remains scoped to the existing authenticated admin-only backend endpoints, and I found no P0 blocker requiring a code change. Current CI is still in progress.\n\n— Robo Bill

PR #1173, opened 2026-09-11
fix: harden Trino registry-only routes and fixture readiness

benben:
## Summary

- Runtime fix: return JSON 404 for unregistered `/bundles` and `/bundles/` API paths instead of serving the admin SPA. Registered bundle authentication and ordinary UI routes are unchanged.
- Wait for the exact expected cell registry after control-plane mode changes in the isolated Trino fixture.
- Retry transient connection failures and stale Service endpoints with bounded requests. Fail if the registry stays wrong or unavailable.
- Check the expected tenant password file with `test -r` on every ready legacy or registered worker before initial writes. Scope permissions to the existing isolated fixture namespaces.
- Preserve all registered-query, ownership and legacy-compatibility assertions. No production configuration or provisioning-readiness changes; document that coordinator reconciliation does not guarantee worker Secret-volume convergence.

## Validation

- The previous live run passed legacy compatibility, initial placement, tenant isolation and green hydration. Its next request raced Service convergence after a control-plane restart.
- Four behavioral regressions exercise the actual shell helper: connection refusal, stale endpoint, persistent wrong registry and unavailable API.
- Clean runs exposed worker Secret-volume lag before both legacy and new-cell initial writes. Twenty-four worker-gate cases cover all-worker checks, three-worker convergence, missing files, stale readiness, cross-cell identity rejection, unavailable pods, denied access and redacted diagnostics. Writes are never retried.
- The next live gate timed out safely before initial writes. In-harness comparison proved the pinned client's request-timeout flag disables its implicit in-cluster configuration. External five-second process bounds preserve working credential discovery; redacted stage diagnostics remain for future failures.
- The shared gate now covers both legacy tenants and the registered tenant. The complete live registry-only proof remains pending.
- The latest live run passed legacy compatibility, registered-cell writes, tenant isolation, green hydration and registered queries after registry-only startup. It exposed the SPA fallback returning HTML 200 for the absent legacy bundle. Composed-router regressions now cover this source correction.
- `just test-mw-fixtures` and `just lint` pass. Independent adversarial review approved the bounded fix.
- Live registry-only validation must pass before this follow-up is ready to merge.

### DIFF
--- a/docs/trino-cells.md
+++ b/docs/trino-cells.md
@@ -107,6 +107,11 @@ or live observer polls. A stopped green does not make blue's tenants unhealthy.
 When green starts, update the registry and restart the control plane. Every
 running backend must reconcile successfully before a tenant is reported ready;
 one successful coordinator cannot conceal another's missing catalog or failure.
+This state confirms control-plane and coordinator catalog reconciliation, not
+tenant-password volume convergence on every worker. A worker can briefly lack
+a newly added tenant's password file even after the coordinator reports ready.
+Initial testing must check worker file availability before the first write;
+do not blindly retry writes when their commit outcome is unknown.
 The console observes the configured routing-active backend. Usage collection
 polls each running backend independently under the existing leader lease.
 
--- a/tests/mw-dev/README.md
+++ b/tests/mw-dev/README.md
@@ -542,6 +542,14 @@ queries the same data through green's independently hydrated catalog. Direct
 coordinator URLs are fixture-only; this does not test Gateway routing or a
 maintenance move of an existing warehouse.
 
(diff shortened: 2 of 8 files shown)


bill-ph (approved):
Reviewed the bounded registry-convergence helper and its behavioral coverage. It retries transient API/refused/stale responses with a finite timeout, validates the exact expected cell set, and fails with an actionable message rather than proceeding on stale configuration. No runtime or deployment configuration changes are introduced, and I found no P0 blocker requiring a code change. The PR remains draft pending the live registry-only validation.\n\n— Robo Bill

PR #1195, opened 2026-09-17
fix(trino): cancel provisioner statements abandoned at the reconcile deadline

fuziontech:
## Why

Tonight (2026-09-17, 02:21 UTC onward) the legacy Trino cell in mw-prod-us stopped reconciling for about 50 minutes. Every control-plane replica logged `reconcile catalogs: list trino catalogs: QUERY_QUEUE_FULL … "root.admin.__admin_provisioner"` or `get nextUri: context deadline exceeded`.

Cause:

- After a coordinator restart, `SHOW CATALOGS` ran longer than the 30s reconcile budget.
- `drainStatement` gave up on the context deadline but never cancelled the query. Trino keeps an abandoned query alive until `query.client.timeout`.
- Those queries had finished executing (100%, root task `FLUSHING`) and were only waiting for a client that had left. They held all 4 running slots in the provisioner's resource group.
- With 6 replicas reconciling, each new tick queued behind them, timed out, and left another abandoned query. The loop can't recover on its own.

Mitigation already applied: I killed the 24 abandoned provisioner `SHOW CATALOGS` queries as the observer principal. Reconcile failures went to 0 on all 6 replicas within the next 90s.

## What

- When the drain stops before Trino returns a terminal response, it sends a best-effort `DELETE` on the pending `nextUri`. This covers context expiry, transport errors, and non-2xx responses.
- The `DELETE` runs on a context detached from the expired reconcile context, bounded to 5s. A failed cancel only logs a warning.
- Statements that already finished or failed aren't cancelled.
- The provisioner credential/source headers are factored into `setStatementHeaders`.

## Testing

- `TestTrinoStatementDrainCancelsAbandonedStatement` uses a fake coordinator whose statement never leaves the queue. It asserts exactly one authenticated `DELETE` on the pending `nextUri`.
  - Mutation-checked: with the cancel call removed, the test fails.
- `TestTrinoStatementDrainDoesNotCancelTerminalStatements` covers finished and failed statements.
- `go test -tags kubernetes ./controlplane/provisioner/` passes; `golangci-lint`: 0 issues.
- **e2e:** not asserted in the Trino lane, because the in-cluster Job can't reliably make a coordinator hold `SHOW CATALOGS` past the reconcile deadline. `tests/mw-dev/README.md` records this and names the unit test that covers it.


benben (approved):

**Approve.** Adds a best-effort DELETE on the pending nextUri when the legacy provisioner client's statement drain exits early, so abandoned SHOW CATALOGS no longer pin the admin resource group. The change is correct for trinoCatalogHTTPClient, but the registered-cell trinoSharedCatalogHTTPClient has the same 30s drain loop with no cancel and is left with the incident's failure mode.

- **major** `controlplane/provisioner/trino_shared_catalog_client.go:109` Shared-catalog client still abandons statements without cancelling on deadline. Registered cell coordinator holds SELECT FROM system.metadata.catalogs past the 30s runStatement timeout.
- **minor** `controlplane/provisioner/trino_provisioner.go:2926` Statement created by POST not cancelled when drain aborts at hop zero. Context expires or initial body fails to parse before pending is assigned from the POST response's nextUri.
- **minor** `controlplane/provisioner/trino_provisioner.go:2980` Non-2xx cancel response swallowed silently, contrary to description. DELETE returns 401 after admin credential rotation or 404; no warning is logged.
- **nit** `tests/mw-dev/README.md:190` Cancel note inserted mid-paragraph about host-qualified user domains. Sentence appears between the Host-header sentence and the Trino worker sizing text.

_Model: fable, PR authored with opus._

PR #1198, opened 2026-09-17
fix(trino): render valid connector names for shared catalogs

benben:
## Summary

Fix shared-catalog provisioning to emit a validated, unquoted connector name in `CREATE CATALOG ... USING ducklake`.

The shared client previously quoted the connector as an SQL identifier. Trino's catalog task passes the rendered identifier to `ConnectorName`, which rejects the embedded quotes. The static catalog client already uses an unquoted connector.

- Validate connector names against Trino's lowercase connector-name grammar before sending a request.
- Keep catalog identifier quoting and property-value escaping unchanged.
- Add regression tests for supported connector names and invalid or injection-like inputs.
- Strengthen the optional shared-catalog end-to-end lane to check the persisted connector and read the newly provisioned catalog.

## Safety and rollout

This change does not clear existing lifecycle ownership or mutation intents, retry ambiguous DDL, or change routing. A previously held intent still requires separate, reviewed recovery. Existing catalogs do not need to be recreated.

## Validation

- TDD: new connector-rendering and invalid-input tests failed against the original client, then passed with the fix.
- `just test-trino 'Trino|SharedCatalog'`: passed (117 top-level tests across four packages).
- `just lint`: passed, zero issues.
- POSIX shell syntax and `git diff --check`: passed.
- `just test-mw-fixtures` with OpenSSL 3 on `PATH`: passed. The macOS system LibreSSL fixture-rendering failure also reproduces on the unchanged base.

The real-cluster shared-catalog lane requires deployment and has not been run for this patch. The retained assertions are not a claim of live rollout validation.


bill-ph (approved):
Reviewed commit 1771ca6f31fcab366f7a59c27000ad6193c571f2. The shared-catalog client now validates connector names before issuing SQL and emits the connector unquoted while preserving catalog identifier quoting and property escaping. Regression tests cover valid names, invalid/injection-like inputs, and the end-to-end assertions verify persisted connector metadata and catalog reads. No P0 blockers found.

Non-blocking note: the live-cluster shared-catalog lane remains deployment-dependent; the retained assertions are appropriately documented as not having run for this patch.

— Robo Bill

PR #1199, opened 2026-09-17
chore(build): bump DuckLake extension to v1.0-posthog.8

fuziontech:
Bumps `DUCKLAKE_EXTENSION_TAG` `v1.0-posthog.7` → `v1.0-posthog.8` in both Dockerfiles.

New in the extension: the batched table-cardinality path with per-snapshot negative caching (PostHog/ducklake#43) and its regression test (PostHog/ducklake#44). `duckdb_tables()`/`information_schema.tables` listings now load `record_count` for all tables in one metadata query instead of one column-stats join per table, and tables with no stats row no longer re-trigger the full-catalog query per lookup. This is the fix for the multi-second catalog listings that trip clients with idle-in-transaction caps.

**Draft until the v1.0-posthog.8 release assets exist** — the Docker build curls them from the GitHub release, which the tag-push pipeline (PostHog/ducklake MainDistributionPipeline) is currently building.


bill-ph (approved):
Reviewed commit c098a77eab75ef76bff58fe438d8ec4723673614. The change is limited to bumping the DuckLake extension tag in both runtime Dockerfiles, consistently bringing in the batched cardinality/negative-cache fix from DuckLake v1.0-posthog.8. No P0 blockers found.

Non-blocking status: this PR is intentionally draft until the v1.0-posthog.8 release assets are available; the build and unit/analysis checks are still running.

— Robo Bill

PR #1200, opened 2026-09-18
fix(duckdbservice): wipe user-attached catalogs at session create on shared-warm workers

fuziontech:
## Problem

DuckDB `ATTACH` is instance-global, and a hot-idle worker pod is reused across sessions of an org. `CreateSession` wiped user **secrets** at session create (`wipeUserSecrets`) but never detached user **catalogs** — so a catalog attached by one session (e.g. an external Postgres source, whose postgres_scanner pool holds live, authenticated upstream connections) was silently inherited by the next session on that worker.

That is the same cross-user isolation boundary as the secret wipe, one level up. It also has a second-order correctness effect that surfaced in production: postgres_scanner's `pg_pool_max_connections` SET has no set-callback and is only read when a catalog's pool is *created* — a session that inherits a stale attach (`ATTACH IF NOT EXISTS` no-ops) cannot reconfigure the pool, the parallel-scan `pg_export_snapshot`/`SET TRANSACTION SNAPSHOT` dance resumes across a load-balanced multi-replica Postgres frontend, and imports intermittently fail with `snapshot "..." does not exist`.

## Fix

`CreateSession` (shared-warm mode, same `max_sessions=1` guard as the secret wipe) now detaches every non-internal attached catalog before the secret wipe, preserving the system-managed reserved set (`ducklake`, `delta`, `memory` — activation and the pg_catalog compat layer own those). A wipe failure fails the session, exactly like secrets. The catalog wipe runs first because attached pools can reference the secrets about to be dropped.

## Tests

- `duckdbservice/user_catalogs_test.go` — unit: wipe detaches user catalogs (incl. quote-unsafe names), preserves reserved/internal ones; wiring: a catalog attached by session A is gone after session B's `CreateSession` on the same worker (red-green verified against the unwired code).
- `tests/mw-dev/e2e/harness.sh` — new `user_catalog_wipe` assertion: attach visible in-session, absent on the next fresh session, `ducklake`+`memory` preserved.
- CLAUDE.md wipe-boundary invariant updated.

Co-authored-by: Shelley <shelley@exe.dev>


bill-ph (approved):
Reviewed commit a6e104a9ca30c135f5ebe094b17edeb19c9219df.

No P0 blockers found. The shared-warm session boundary now detaches user catalogs before wiping secrets, preserves reserved/internal catalogs, quotes user-controlled identifiers safely, fails session creation on wipe errors, and adds unit, wiring, and mw-dev E2E coverage.

— Robo Bill

benben (approved):

**Approve.** Adds a session-create DETACH of every non-internal, non-reserved catalog on shared-warm workers, mirroring the secret wipe. The mechanism and tests are sound for user attaches, but the reserved set does not name DuckLake's Postgres-typed metadata catalog (`__ducklake_metadata_ducklake`); if duckdb_databases() does not flag it internal, the wipe detaches the org's own metadata catalog on every fresh session, and the unit tests (all `:memory:` attaches) cannot detect that.

- **major** `duckdbservice/user_catalogs.go:48` DuckLake metadata catalog not reserved; wiped if not flagged internal. Confirm `SELECT internal FROM duckdb_databases() WHERE database_name='__ducklake_metadata_ducklake'` is true on a real worker; otherwise every CreateSession detaches it and the reaper-configured pg pool from configureDuckLakeMetadataPool is lost.
- **minor** `duckdbservice/service.go:1472` DestroySession still leaves user catalogs (live upstream pools) on hot-idle worker. Secrets get a best-effort wipe at destroy but attached postgres_scanner pools stay open for the whole hot-idle TTL.
- **minor** `duckdbservice/user_catalogs_test.go:48` Tests only attach :memory: catalogs; no real DuckLake/type catalog covered. A regression where a system-owned non-internal catalog gets detached passes both unit tests.

_Model: fable._

benben (approved):

**Approve (re-review of new commits).** The new commit only reworks the harness's same-session check to ride one psql connection; that part is correct, but the "fresh session" leak assertion still cannot land on the worker holding the attach under the e2e's exploratory tier.

- **major** `tests/mw-dev/e2e/harness.sh:2582` Leak check passes trivially: fresh SELECT never reuses the pinned worker. mw-dev sets DUCKGRES_EXPLORATORY_TIER_ENABLED=true and findIdleAssignedWorkerLocked matches on profile, so the ATTACH (DDL, pinned standard) and the plain SELECT (exploratory) land on different-shaped workers.
- **minor** `tests/mw-dev/e2e/harness.sh:2579` Inline psql bypasses pg_script backpressure retry; hard-fails on transient errors. A 'no Duckgres worker'/'capacity exhausted' response on the attach connection aborts the whole harness instead of retrying like every other assertion.

_Model: fable._

fuziontech:
Superseded by #1201 (merged as 261f4304), which lands the same session-boundary catalog detach. #1201 adopted this PR's single-connection harness technique, run-unique catalog name and system-catalogs-preserved assertion, and additionally resolves the primary database by oid rather than reserving `memory` by name, catalog-qualifies the listing, gives the detach its own deadline, and detaches best-effort at session destroy.

PR #1205, opened 2026-09-19
feat(trino): manage shared compute pools with fenced admission

benben:
## Summary

- Add an opt-in controller for a shared pool of immutable Trino clusters. Keep existing provisioning and routing unchanged when disabled.
- Publish shared catalog definitions through a fenced writer. Admit tenants only after serving coordinators acknowledge the required catalog and authentication configuration.
- Reconcile cluster creation, validation, replacement, drain and retirement through the Gateway lifecycle API. Preserve pinned work during planned replacement; treat proven coordinator loss separately.
- Persist operation identities and retry state so response loss, leadership changes and repeated tenant changes can recover without replaying an obsolete operation.
- Include unit, PostgreSQL and protocol-contract regressions, plus opt-in deployment acceptance tests.

## Validation

All of the following ran on the final integrated tree:

- `go build -tags kubernetes ./...` and `go build ./...`
- `go test -tags kubernetes -count=1 ./controlplane ./controlplane/provisioner ./controlplane/provisioner/opa ./controlplane/configstore ./controlplane/trinocatalog ./controlplane/trinogateway ./controlplane/trinopool` — pass
- `go test -count=1 ./server/ ./transpiler/...` — pass
- Real PostgreSQL: `go test -tags kubernetes -count=1 ./tests/configstore/ ./tests/trinocatalog/` — pass (migrations, fenced writer and publication records against a live database)
- `go test -race -tags kubernetes` over the publication, admission and member-lifecycle cases — pass
- `golangci-lint run` — 0 issues

Not run here: the Docker-backed suites (`controlplane/admin`'s container tests
and `tests/integration/`), which need a Docker daemon this environment does not
have. Live pool deployment and zero-downtime behaviour are unverified and
remain separate checks — nothing in this PR exercises a real Gateway, a real
pooled coordinator, or a rollout.

Two behaviours are deliberately conservative and documented in the contract
rather than hidden: a tenant admission is held (not silently advanced) while an
earlier publication of the same occurrence may still be in flight, and a
suspected member always leaves through drain or a proven-terminated claim
rather than being locally recovered.

Known limitation, not an automatic recovery: a hold is released once the
request it is waiting on settles, but one case does not clear on its own. If a
tenant's first publication has an UNKNOWN outcome and the retry of that same
request is then definitively refused for a condition that never goes away, the
tenant stays held — it publishes no corrected binding and is not revoked until
that refusal stops, which may need operator intervention. Its already-admitted
principals keep working; the pending change does not take effect. This is the
conservative direction: releasing the hold could let the earlier in-flight copy
overwrite a newer binding with no way to detect it. Resolving it properly needs
either a durable record of whether an occurrence was ever uncertain, or a way to
retire a request identity at the Gateway, and both are follow-ups rather than
part of this change.

## Catalog store identity

The publisher requires an explicit `DUCKGRES_TRINO_POOL_CATALOG_CELL_ID`, matching the coordinators’ `catalog-store.cell-id`. It is separate from the durable pool/warehouse-assignment ID, which still addresses publication checkpoints. Missing, blank or invalid settings fail before the database opens. The publisher bootstraps an empty configured schema and reconciles warehouse definitions without copying an old catalog partition.

Fresh-store partition/checkpoint regressions failed before the fix and pass against disposable PostgreSQL. The control-plane and catalog suites pass; `just lint` reports zero issues. An additional Kubernetes-tagged lint run reports pre-existing findings outside the files changed by this identity update. Docker-backed and live deployment tests remain unavailable/unverified as described above.

## Rollout

The feature is disabled by default. This PR does not activate or deploy a pool. Activation requires compatible Gateway and Trino images, database grants and deployment configuration. Live deployment and zero-downtime acceptance remain separate checks.

## Agent context

Implemented and reviewed with Claude Code agents in isolated worktrees; Codex coordinated integration and verification.


bill-ph (approved):
Focused static review of 0db3cb2579220c578349683efa262cbe79534a55. No P0 blocker identified; approving under the requested P0-only approval policy, with the following actionable findings:

1. **[P1] Include tenant-admission state in the configure step identity** — `controlplane/trino_pool_operator.go:441,459,467-476`. Changing only `TenantAdmission` during a leadership term changes the ConfigurePool request body but not its operation/step ID. Under the documented Gateway whole-body idempotency contract this conflicts with the previous request. ConfigurePool runs before tenant and instance progress, so repeated conflicts prevent the new gate setting and stop reconciliation for that term. Include all request-shaping fields in the durable intent and add a test with a replay/hash-enforcing fake (the current ConfigurePool fake does not enforce this).

2. **[P2] Give each configuration transition its own durable identity** — `controlplane/trino_pool_operator.go:429-441`. A same-epoch A → B → A change (for example changing MaxSurge and then restoring it) reuses A's already completed step. An idempotent Gateway replays A's original result rather than applying A over B, and this caller discards the returned pool state. The Gateway can therefore remain on B while the database/controller desire A. Persist a monotonically advancing configuration intent/version, retaining its step ID only for retries of that occurrence; test A → B → A against a replaying Gateway.

3. **[P2] Reconcile excess current-release instances on scale-down** — `controlplane/trinopool/plan.go`, the `outdatedServing` / empty-result branch in `PlanNext`. With three SERVING instances on the desired release, DesiredInstances=2 and MinServing=2, there is no deficit, preparing member, drain, or outdated member, so the planner returns “pool matches the desired release and instance count” indefinitely. No member enters the safe drain/retire path, leaving the excess compute running. Add an excess-capacity drain decision respecting the floor and one-operation limit, plus a same-release scale-down test.

Scope/overengineering check: the feature remains opt-in, and the durable fencing/retirement machinery is relevant to the stated safety requirements. No separate scope-creep blocker identified. The configuration identity problems can be addressed in the existing durable-intent model rather than introducing another reconciliation subsystem.

Validation: inspected controller configuration, lifecycle/planning, Gateway client contract, admission/publication and resource-deletion paths and relevant tests. This is a focused review, not an exhaustive audit of the 22k-line addition. I did not run tests or a live deployment. Current CI reports unit/controlplane/configstore/Kubernetes tests successful; integration/e2e work was still in progress when checked. The author-documented UNKNOWN-publication hold remains an operational limitation, not a newly discovered finding.

— Robo Bill

PR #1206, opened 2026-09-20
fix: separate shared-pool auth from fixed rollout readiness

benben:
## Summary

Shared-pool Gateway authentication reuses `DUCKGRES_TRINO_ROLLOUT_TOKEN_FILE`. The fixed-slot readiness factory incorrectly treated that token as opting into static canary validation, preventing pool-only configurations from starting.

- Select fixed-cell readiness by the actual registry mode.
- Leave readiness disabled for pool-only registries without explicit canary configuration, including deployments retaining the legacy single coordinator.
- Keep incomplete fixed/mixed configuration and explicit invalid canaries fail-closed. Pool Gateway token validation remains unchanged.
- Document the configuration boundary and real-cluster acceptance limitation.

## Validation

- Red regression reproduced the startup failure with a real shared-pool mode fixture; all 15 configuration cases now pass.
- Existing readiness, managed Gateway, and pool wiring tests pass through `just test-trino`.
- `just lint` passes with an isolated cache; `git diff --check` passes.
- Independent adversarial review completed.
- Real-cluster pool acceptance still requires deployment of the resulting image. The existing active-pool harness covers that path; this PR does not claim it has passed.


bill-ph (approved):
Reviewed 86e01189. No actionable findings or P0 blocker identified.

The readiness factory now excludes shared-pool cells from fixed-slot validation while preserving failure for incomplete fixed/mixed configurations and explicit pool-only canary configuration. Legacy unregistered wiring does not accidentally count as a fixed cell. I cross-checked the separate pool Gateway construction/token-reading path; this change does not remove that path. The table-driven tests cover the important selection boundaries and verify that mixed-fleet handlers contain only fixed slots.

Overengineering/scope-creep check: the small mode-based filter addresses the shared-token coupling without adding a new flag or credential. The test and documentation additions are proportional and in scope.

Validation: static review of the full diff and surrounding readiness and pool Gateway wiring. I did not run local tests or real-cluster acceptance; the documented deployment/active-pool acceptance check remains necessary.

— Robo Bill

